PR #25, opened 2015-06-12
Change key for measurement in input payload

kvikas:
It seems key name in input payload to add points is changed to 'measurement' from 'name'.

### DIFF
--- a/content/docs/v0.9/concepts/reading_and_writing_data.md
+++ b/content/docs/v0.9/concepts/reading_and_writing_data.md
@@ -76,7 +76,7 @@ Which returns:
         {
             "series": [
                 {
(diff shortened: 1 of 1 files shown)


otoolep:
Thanks @kvikas -- we're in the middle of overhauling the docs right now, so I'll leave it to @gunnaraasen to decide if this should be merged.

gunnaraasen:
@kvikas thanks for the PR, but the current doc is correct in this case. Only the JSON write API switched from `name` to `measurement`.

The JSON in the doc shows the result of a query, which is returned by InfluxDB in a generic wrapper schema. Since queries can return results other than than points (like a list of databases), the generic `name` is a better fit and reflects what InfluxDB actually returns.

Issue #20, opened 2015-06-11
Use actual version number in RPM filename instead of "latest"


labrown:
Instructions on http://influxdb.com/download/index.html show the RPM file name `influxdb-latest-1.x86_64.rpm` which doesn't work with Ansible at all.  My playbook skips right over installing that RPM.

I could not find an `influxdb-0.9.0-1.x86_64.rpm` on your S3.  Can you please make it available so those of us using change management processes can take advantage?

gunnaraasen:
@labrown We'll post a rpm with v0.9.0 in the filename soon.

gunnaraasen:
Versioned RPM (and deb) download links are now on the [0.9 installation page](http://influxdb.com/docs/v0.9/introduction/installation.html).

PR #31, opened 2015-06-15
Fix install instructions

rmoff:
D/l'd file didn't match name in next statement of instructions


gunnaraasen:
Thanks @rmoff

otoolep:
I don't think this is correct. Soon we'll have a 0.9.1 release. Instead 0.9 should have been replaced with "latest", no?

@gunnaraasen

gunnaraasen:
This is a fix for #20. The downloads page will continue to point to latest and the installation page will need to be updated for every 0.9.x release.

PR #35, opened 2015-06-15
Add default configuration file location

anaptfox:
The docs for v0.9 never say where the default configuration file is located.


beckettsean:
@anaptfox Thanks for the good suggestion. It would be nice to tell everyone where the default file is located, but the path is OS-specific. For instance, on OS X the default configuration file is in `/usr/local/etc/`, not `/etc/opt/influxdb/`. Rather than document every case for every OS we choose to describe how to make a new one, which is the preferred method, as it guarantees compatibility.

Issue #14, opened 2015-06-11
Link to 0.8 docs doesn't change nav links. 


beckettsean:
All the links in the nav on http://influxdb.com/docs/v0.8/introduction/overview.html go to 0.9 docs. There's no way to browse the 0.8 documentation any longer. This is particularly tough since Hosted is still on 0.8

Issue #28, opened 2015-06-12
Add deprecation warning to old docs pages


gunnaraasen:
A bunch of old docs pages are linked to from old tutorials and blog posts. There should be a note at the top of all old docs pages notifying them that 0.9.0 is the latest release.

Issue #26, opened 2015-06-12
0.8 docs cannot be browsed


beckettsean:
Direct URLs still work but there's no way to browse docs or follow breadcrumbs. Apparently Hugo doesn't make it easy.

As a simple workaround I propose the following: 
 - make a TOC page with links to all the 0.8 docs, similar to the old left side nav bar
 - on each 0.8 page have a link at the top and bottom of the page to the TOC page
 - can we remove or disable the right side nav for the 0.8 pages, or make some note that it's invalid?

Issue #21, opened 2015-06-11
Images missing on old blog post


beckettsean:
http://influxdb.com/blog/2014/09/26/one-year-of-influxdb-and-the-road-to-1_0.html

This is worth fixing since we link to it in the second paragraph on the Overview page. All other blog posts look fine.

![screen shot 2015-06-11 at 14 13 46](https://cloud.githubusercontent.com/assets/6537329/8118499/17e22186-1044-11e5-8ab5-910e8bc77b87.png)

PR #36, opened 2015-06-16
Old docs update

gunnaraasen:
With this update, old docs should display correctly.


gunnaraasen:
Fixes #14, #21, #26, #28.

PR #41, opened 2015-06-17
docs on authorization might be incorrect

qdonnellan:
Influx gang - I'm in the process of setting up influxdb locally and have been playing with authentication; it seems that the docs on the matter might be slightly off (or the config file is incorrectly structured, which I'm inclined to not believe).


qdonnellan:
Bah! Just realized I'm PR'ing this from my fork/master, going to create a separate branch... instead

PR #42, opened 2015-06-17
proposed change to the authorization docs

qdonnellan:
Influx gang - I'm in the process of setting up influxdb locally and have been playing with authentication; it seems that the docs on the matter might be slightly off (or the config file is incorrectly structured, which I'm inclined to not believe).


otoolep:
+1, now inline with the source code.

Thanks @qdonnellan

PR #47, opened 2015-06-19
Read/write example: POST payload must not start with new line.

diarized:
In the documentation, in the basic read/write example of writing multiple records at once, if the payload starts with new line, it does not work.


beckettsean:
@diarized great catch, thanks!

diarized:
Thanks

PR #57, opened 2015-06-23
fix typo ("tags" should be "tag")

jeffbmartinez:
Fixing a typo. "tags" should be singular in the following sentence:

"...with the measurement name of cpu and **tags** host has now been written..."

Found on https://influxdb.com/docs/v0.9/introduction/getting_started.html


otoolep:
Thanks @jeffbmartinez

Issue #56, opened 2015-06-23
Bad link in CLI Docs


mjdesa:
In the [CLI](https://influxdb.com/docs/v0.9/clients/shell.html) documentation under the `Issuing Queries` section, the `online documenation` link sends you to [https://influxdb.com/docs](https://influxdb.com/docs)

mjdesa:
Solved by [58](https://github.com/influxdb/influxdb.com/pull/58)

PR #58, opened 2015-06-23
Fix bad link in CLI documentation

mjdesa:
Solved issue [56](https://github.com/influxdb/influxdb.com/issues/56).
Link now leads to `/docs/v0.9/query_language/querying_data.html` instead of `/docs`


gunnaraasen:
:+1:

Issue #53, opened 2015-06-22
Improve `Querying Data`


mjdesa:
The first curl example in [Query Data](https://influxdb.com/docs/v0.9/query_language/querying_data.html)

```bash
curl -G 'http://localhost:8086/query' --data-urlencode "q=SELECT * FROM cpu"
```

returns 

```bash
{"results":[{"error":"database name required"}]}
```

Specifying the database in the command to

```bash
curl -G 'http://localhost:8086/query' --data-urlencode 'q=SELECT * FROM "mydb"."default".cpu'
```

yields a more desirable result.

mjdesa:
Note the change of the `"`s to `'`s following the `--data-urlencode`

Alternatively we could use escaped quotes `\"`s, like so

```bash
curl -G 'http://localhost:8086/query' --data-urlencode "q=SELECT * FROM \"mydb\".\"default\".cpu"
```

mjdesa:
Another alternative is

```bash
curl -G 'http://localhost:8086/query' --data-urlencode "db=mydb" --data-urlencode "q=SELECT * FROM cpu"
```

PR #59, opened 2015-06-23
Change first curl example to specify a database

mjdesa:
As noted in [53](https://github.com/influxdb/influxdb.com/issues/53),
the current query returns `{"results":[{"error":"database name
required"}]}`. To remedy this, I specify a database `mydb`.


gunnaraasen:
+1

PR #80, opened 2015-06-30
Add '/' to deperecated seperators

ecki:
The `/` is even worse than `.` in a regular expression. The text could be even more generic if you prefer?


beckettsean:
@ecki the original text wasn't in reference to regular expressions, but to the fact that any identifiers with `.` must be double-quoted whenever they are referenced. Slashes would also need to be double-quoted, as would anything not `[A-z,0-9,_]`.

We've got a massive docs refactoring happening in the `update-rw-docs` branch. I think I'll remove the text entirely from this page and put it into the syntax guides and the troubleshooting page we are adding.

beckettsean:
@ecki thanks very much for pointing this out, and for showing me that it's easy to conflate InfluxQL's quoting requirements with recommending easy regex strings.

ecki:
Thanks for the quick response and actual interesting answer. I will see if I find more stuff in the new branch which can be improved for newbies to influxDB :)

Issue #103, opened 2015-07-08
Update 0.9 docs for new derivative behavior


nathanielc:
The derivative docs online for 0.9 still say the same thing as 0.8. Since now there is a second argument to the function and it changes the behavior we should update the docs. I'll take a stab at it but this  https://github.com/influxdb/influxdb/issues/2699#issuecomment-110527407 is my only source of information on how it really works so my explanation may not be adequate.

PR #104, opened 2015-07-08
Update 0.9 derivative

nathanielc:
Not sure this explanation is entire correct but based on discussion here https://github.com/influxdb/influxdb/issues/2699# and my own testing this seems to be the new behavior.

This is for issue https://github.com/influxdb/influxdb.com/issues/103

### DIFF
--- a/content/docs/v0.9/query_language/functions.md
+++ b/content/docs/v0.9/query_language/functions.md
@@ -90,7 +90,7 @@ SELECT PERCENTILE(field_key, N) FROM measurement WHERE time > now() - 1d GROUP B
 
 This definition is a work in progress, suggestions welcome.
 
-DERIVATIVE() can have two arguments. The first is required and is a field name. The second is optional and is a rate normalization parameter. The out is a field containing the value of `(v_last - v_first) / (t_last - t_first)` where `v_last` is the last value of the given field and `t_last` is the corresponding timestamp (and similarly for `v_first` and `t_first`). In other words, DERIVATIVE() calculates the **rate of change** of the given field.  The optional second argument determines the time `units` of the output. For example `DERIVATIVE(field_key, 1s)` returns a rate per second while `DERIVATIVE(field_key, 1h)` returns a rate per hour. The second argument can be any time duration and if it is not specified defaults to `1s` or the time duration of a `GROUP BY` statement.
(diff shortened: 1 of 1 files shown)


content/docs/v0.9/query_language/functions.md
```diff
@@ -90,12 +90,27 @@ SELECT PERCENTILE(field_key, N) FROM measurement WHERE time > now() - 1d GROUP B
...
 
-DERIVATIVE() requires exactly one argument, which is a field name. The out is a field containing the value of `(v_last - v_first) / (t_last - t_first)` where `v_last` is the last value of the given field and `t_last` is the corresponding timestamp (and similarly for `v_first` and `t_first`). In other words, DERIVATIVE() calculates the **rate of change** of the given field.
+DERIVATIVE() can have two arguments. The first is required and is a field name. The second is optional and is a rate normalization parameter. The out is a field containing the value of `(v_last - v_first) / (t_last - t_first)` where `v_last` is the last value of the given field and `t_last` is the corresponding timestamp (and similarly for `v_first` and `t_first`). In other words, DERIVATIVE() calculates the **rate of change** of the given field.  The optional second argument determines the time `units` of the output. For example `DERIVATIVE(field_key, 1s)` returns a rate per second while `DERIVATIVE(field_key, 1h)` returns a rate per hour. The second argument can be any time duration and if it is not specified defaults to `1s` or the time duration of a `GROUP BY` statement.
```
gunnaraasen: Small typo: "put" should be "output"

gunnaraasen:
This looks correct to me.

@jwilder can you take a quick look at this?

otoolep:
I think @dgnorton implemented derivative, but I might be mistaken.

jwilder:
Only thing missing is that I see is that you can also use an aggregate function over the value as well such as derivative(mean(value), 1s).

nathanielc:
Thanks for the feedback. Look good now?

jwilder:
:+1:

gunnaraasen:
Thanks @nathanielc!

nathanielc:
@jwilder I am confused again about the derivative behavior. How does the derivative behave with multiple aggregation functions?

I have a measurement (from collectd) called interface and it has tags: host, dsname, role. Where dsname is one of 'rx' or 'tx' for incoming outing network bytes. 

I have this query:

```
SELECT derivative(sum(value), 1s)  FROM "interface" WHERE "role" = 'collectd' AND "host" =~ /.*/ AND time > now() - 1h GROUP BY time(10s), "dsname"
```

I was hoping that it would first `sum` across all hosts, time slices for the specific role and then take a derivative for the first and last data points. I am getting really large negative values in the result which should not be possible. If change the query to group by host and remove the `sum` function I get sane results. Any ideas?

emanuelis:
I have the same problem as @nathanielc 
```
SELECT non_negative_derivative(sum(value)) FROM "interface_rx" WHERE "host" =~ /.*example.com$/ AND "type" = 'if_octets' AND "instance" = 'eth0' AND $timeFilter GROUP BY time($interval)
```
Can not figure out how to aggregate many hosts traffic into one graph.
There are a few problems:
1. The last value always is very small ("non_negative_" solves that).
2. The first value is very high. <- this is the biggest problem
3. When one host's counter resets - the sum becomes wrong.

PR #136, opened 2015-07-21
Rework derivative definition

mjdesa:
Looking for feedback on definition

### DIFF
--- a/content/docs/v0.9/query_language/functions.md
+++ b/content/docs/v0.9/query_language/functions.md
@@ -90,8 +90,6 @@ SELECT PERCENTILE(field_key, N) FROM measurement WHERE time > now() - 1d GROUP B
 
 DERIVATIVE() computes the rate of change for a field value at each point in the series.
 
-This definition is a work in progress, suggestions welcome.
-
 DERIVATIVE() can have two arguments. The first is required and is a field name. The second is optional and is a rate normalization parameter. If the second parameter is not provided defaults to 1s
 
 The optional second argument determines the time `units` of the output. For example `DERIVATIVE(field_key, 1s)` returns a rate per second while `DERIVATIVE(field_key, 1h)` returns a rate per hour.
--- a/content/docs/v0.9/query_language/functions.md
+++ b/content/docs/v0.9/query_language/functions.md
@@ -88,10 +88,13 @@ SELECT PERCENTILE(field_key, N) FROM measurement WHERE time > now() - 1d GROUP B
 
 ## Derivative
 
+DERIVATIVE() computes the rate of change for a field value at each point in the series.
+
 This definition is a work in progress, suggestions welcome.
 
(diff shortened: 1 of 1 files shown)


mjdesa:
@beckettsean @gunnaraasen  feedback?

content/docs/v0.9/query_language/functions.md
```diff
@@ -88,10 +88,13 @@ SELECT PERCENTILE(field_key, N) FROM measurement WHERE time > now() - 1d GROUP B
 
 ## Derivative
 
+DERIVATIVE() computes the rate of change for a field value at each point in the series.
+
 This definition is a work in progress, suggestions welcome.
 
-DERIVATIVE() can have two arguments. The first is required and is a field name. The second is optional and is a rate normalization parameter. The output is a field containing the value of `(v_last - v_first) / (t_last - t_first)` where `v_last` is the last value of the given field and `t_last` is the corresponding timestamp (and similarly for `v_first` and `t_first`). In other words, DERIVATIVE() calculates the **rate of change** of the given field.  The optional second argument determines the time `units` of the output. For example `DERIVATIVE(field_key, 1s)` returns a rate per second while `DERIVATIVE(field_key, 1h)` returns a rate per hour. The second argument can be any time duration and if it is not specified defaults to `1s` or the time duration of a `GROUP BY` statement.
+DERIVATIVE() can have two arguments. The first is required and is a field name. The second is optional and is a rate normalization parameter and if not parameter is provided defaults to 1s.
```
gunnaraasen: Last sentence is a run on sentence. The `defaults to 1s` seems more appropriate for the next paragraph.

mjdesa: kk

gunnaraasen: Is there a reason the explanation of how the derivative function works is being removed. If anything, I feel like we should be expanding the explanation instead of cutting it out completely.

mjdesa: The definition that we have there isn't entirely true.

mjdesa: The units `t_first` and `t_last` fields are expressed in can vary and will yield different results.

I agree the definition should be expanded, but I don't know that this is the place for an in depth explanation.

gunnaraasen: Where should the full explanation live? The functions page seems as good a place as any for a full description.

I know you did some testing on the type results for `derivative`. Is that something we can add to the explanation?

mjdesa: I suppose. I guess it just feels like the derivative section will be substantially longer than any other on the page.

I'll add an example.

beckettsean:
I think this works for now, since any additional explanation needs to add to my understanding, and the old text didn't make anything more clear to me. Let's start here and find out what people need to know more about.

PR #138, opened 2015-07-22
show retention policies now requires "ON"

tmonk42:
Was playing with the latest nightly build, ran into an issue trying to show retention policies. Looks like syntax got updated recently. https://github.com/influxdb/influxdb/commit/ee8ba11c4fdb30f5ae5c7d1c468d7985bf2232fd

Looks like this syntax does NOT work with 0.9.1. Not sure what the style guideline would be here for denoting differences between minor versions. I'll gladly update this PR if there's an example of how to show 0.9.1 syntax vs 0.9.2.

Thanks!

### DIFF
--- a/content/docs/v0.9/administration/administration.md
+++ b/content/docs/v0.9/administration/administration.md
@@ -151,13 +151,13 @@ Durations such as `1h`, `90m`, `12h`, `7d`, and `4w`, are all supported and mean
 To delete a retention policy issue the following command:
 
 ```sql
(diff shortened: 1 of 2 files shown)


beckettsean:
thanks @tmonk42 this was a recent intentional change to the syntax to make it congruent with the other SHOW statements.

We had planned to simply update the page with the push to 0.9.2 and make a note in the changelog about the switch, as well as a small note to the mailing list.

PR #144, opened 2015-07-23
Add 0.9.2 release announcement

pauldix:
it's not released yet, but will be in a few hours. For when that time comes...


beckettsean:
LGTM. Can't think of anything else we need to mention here, unless the HTTPS option rises to that level of meaning.

PR #163, opened 2015-07-28
Add Google site verification meta tag

gunnaraasen:


beckettsean:
presumably the code is something we generated/received from Google? LGTM

gunnaraasen:
Got it from @toddboom so I assume it's legit. :)

PR #165, opened 2015-07-28
Use correct case insensitive regex flag syntax

labrown:
Use the correct regex flag syntax for case insensitive matching.


beckettsean:
Thanks, @labrown, great catch.

PR #183, opened 2015-07-30
Updated client library page to reflect official php libary

thecodeassassin:


content/docs/v0.9/clients/api.md
```diff
@@ -27,6 +27,7 @@ This is a list of the client libraries which currently support InfluxDB version
 
 ## PHP
 
+- **Official** [InfluxDB-PHP Github](https://github.com/influxdb/influxdb-php) by [InfluxDB](https://github.com/influxdb)
```
beckettsean: @pauldix @toddboom we have no other libraries declared "Official". Is there criteria for such? Does the influxdb-php repo meet the informal criteria? Are there others on this page (Java, Ruby, Rails) that we should also label as "Official"

pauldix: hmmm. The only one I would absolutely call official is the Go library. I guess we should point people at what we think is the best one. For the PHP one I think that's the one if it's the new one that codeassasin has been working on.

beckettsean: How about "Official" for InfluxDB-maintained libraries, which currently means only the Go library. Then "Recommended" for most favored versions per language? Maybe Recommended implies too much testing/compatibility on our part? Maybe "Preferred" or "Primary" or some neat BttF reference I can't think of right now?

beckettsean:
@thecodeassassin until we have an official certification regime we aren't comfortable recommending or blessing a particular repo over another.

beckettsean:
Added in a direct commit without the "Official" language.

thecodeassassin:
Thanks @beckettsean

Issue #115, opened 2015-07-14
new page: document how InfluxDB is like a SQL database, how it is not


beckettsean:
E.g. measurements are like tables, tags are like indexed columns, fields are like unindexed columns, points are like rows. CQs are a bit like stored procedures or triggers, as are replication policies. 

However, schema is not pre-defined. Points can have one to all of the fields for a measurement, zero to all tags. Indicies are determined by the system and not the user. Measurements can't be JOINed.

beckettsean:
closed by #187

PR #187, opened 2015-08-01
Add sql to influxdb comparison page [WIP]

rkuchan:
Created a rough draft of the SQL to InfluxDB comparison page.


rkuchan:
https://github.com/influxdb/influxdb.com/issues/115

rkuchan:
@mjdesa @beckettsean @gunnaraasen Please help!

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+
+Other concerns are italicized in the text.
+
+# What's in a database?
+
+This page gives SQL users an overview of how InfluxDB is like an SQL database and how it's not. It highlights some of the major distinctions between the two and provides a loose crosswalk between the different database terminologies and query languages.
+
+## In general...
+
+InfluxDB is designed to work with time-series data. SQL databases can handle time-series but weren't created strictly for that purpose. In short, InfluxDB is made to store a large volume of time-series data and perform real-time analysis on those data, fast.
```
beckettsean: nit: replace "fast" with "quickly"

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+
+## In general...
+
+InfluxDB is designed to work with time-series data. SQL databases can handle time-series but weren't created strictly for that purpose. In short, InfluxDB is made to store a large volume of time-series data and perform real-time analysis on those data, fast. 
+
+### Timing is everything
+
+In InfluxDB, a timestamp identifies a single point in any given data series. This is like an SQL database table where the primary key is pre-set by the system and is always time.
+
+InfluxDB also recognizes that your schema preferences may change over time. In InfluxDB you don't have to define schemas up front. Data points can have one to all of the fields for a measurement and zero to all tags, and you can add fields to existing measurement-tag combinations without penalty. If you need an explanation of the terms measurements, tags, and fields check out the next section for an SQL database to InfluxDB terminology crosswalk.
```
beckettsean: "Data points can have one to all of the fields for a measurement and zero to all tags," 

the "one to all" phrasing feels awkward. Maybe it's more than one sentence. 

E.g.

Data points can have only one of the fields on a measurement or all of the fields on a measurement, or any number in between. New fields can be added to a measurement simply by writing a point with the new field.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+
+## Terminology
+
+The table below is a (very) simple example of a table  called `foodships` in an SQL database with the unindexed column `no_foodships` and the indexed columns `id`, `planet`, and `time`.
```
beckettsean: I think "no_foodships" would be clearer as `#_foodships` or just `foodships`. I parse it first as "no foodships", not "number of foodships", which makes no sense in this context.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+|    3 | Jupiter | 1429185602000000000 |           21 |
+|    3 | Jupiter | 1429185603000000000 |           20 |
+|    4 | Saturn  | 1429185600000000000 |            5 |
+|    4 | Saturn  | 1429185601000000000 |            5 |
+|    4 | Saturn  | 1429185602000000000 |            6 |
+|    4 | Saturn  | 1429185603000000000 |            5 |
++------+---------+---------------------+--------------+
+```
+
+*I'm guessing you'd use a different SQL database setup for these data. How upset will people be by this? Is this too childlike?*
```
beckettsean: This is great! Humor is useful in a doc like this. The biggest problem is someone will complain that the times don't match up with actual orbits.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+Referencing the example above, in general:
+
+* An InfluxDB measurement (`foodships`) is similar to an SQL database table.
+* InfluxDB tags ( `id` and `planet`) are like indexed columns in an SQL database.
+* InfluxDB fields (`no_foodships`) are like unindexed columns in an SQL database.
+* InfluxDB points (for example, `2015-04-16T12:00:00Z	5`) are similar to SQL rows.
+
+Building on this comparison of database terminology, InfluxDB's [continuous queries](https://influxdb.com/docs/v0.9/query_language/continuous_queries.html) and [replication policies](https://influxdb.com/docs/v0.9/administration/administration.html) are similar to stored procedures in an SQL database. They're *specified once and then performed regularly and automatically.*
+
+*Should I mention triggers? From what I can tell pre-coded triggers run when something specific happens to a column but I don't have any experience with triggers.*
```
beckettsean: Nothing we have is like a trigger. I'd leave it alone for now, or we could say that there's nothing analogous to triggers in InfluxDB.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+
+Of course, there are some major disparities between SQL databases and InfluxDB. SQL JOINs aren't available for InfluxDB measurements; your schema design will reflect that difference. And, as we mentioned above, a measurement is like an SQL table where the primary index is always pre-set to time. InfluxDB timestamps must be in UNIX epoch (GMT) or formatted as a date-time string valid under RFC 3339. 
+
+*Should there be something about sequential processing here?*
```
beckettsean: ?? I don't know what you mean by sequential processing. The words I understand but in context I can't gather the meaning. Are they jargon of some kind?

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+For more detailed descriptions of the InfluxDB terms mentioned in this section see our [Glossary of Terms](https://influxdb.com/docs/v0.9/concepts/glossary.html).
+	
+## InfluxQL and SQL
+
+InfluxQL is an SQL-like query language for interacting with InfluxDB. It has been lovingly crafted to feel familiar to those coming from other SQL or SQL-like environments while also providing features specific to storing and analyzing time series data.
+
+InfluxQL's SELECT statement follows the form of an SQL SELECT statement:
+
+```sql
+SELECT whatever FROM measurement_name WHERE some_conditions
```
beckettsean: nit: I would replace "whatever" with "stuff", since we actually don't support very many things in the SELECT clause.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -0,0 +1,154 @@
...
+If you wanted to see data for the planet Saturn after 12:00:01 GMT on April 16, 2015, you'd enter:
+
+```sql
+SELECT * FROM foodships WHERE planet = 'Saturn' and time > '2015-04-16 12:00:01'
+```
+
+As shown in the example above, InfluxQL allows you to specify the time range of your query in the WHERE clause. You can use date-time strings wrapped in single quotes that have the format `YYYY-MM-DD HH:MM:SS.mmm` ( `mmm` is milliseconds and is optional). You can also use relative time with `now()` which refers to the server's current timestamp:
+
+```sql
+SELECT * FROM foodships WHERE time < now() -1h
```
beckettsean: This is a legal but unadvisable query. Without an explicit lower bound the time range starts at epoch 0. This would query for everything from 1-1-1970 until an hour ago. `WHERE time > now() -1h` is everything from the past hour, which is a more common query type for InfluxDB use cases.

beckettsean:
Great start @rkuchan! A few edits, and we need some input from @gunnaraasen and @mjdesa and maybe @toddboom or @corylanou? but very close to ready to merge.

beckettsean:
merging so we can push to new.influxdb.com

PR #195, opened 2015-08-04
fix: Specify "yum" not "rpm"

brianredbeard:
When installing packages on RPM based systems the best practice
is to use the command "yum localinstall" not "rpm".  This mananges
any dependencies on behalf of the user.  While influxdb should be
self sufficient, it's best to rely on the package manager and not
it's primitives.

### DIFF
--- a/content/docs/v0.6/introduction/installation.md
+++ b/content/docs/v0.6/introduction/installation.md
@@ -30,11 +30,11 @@ RedHat and CentOS users can install by downloading and installing the rpm like t
 ```bash
 # for 64-bit systems
 wget https://s3.amazonaws.com/influxdb/influxdb-0.6.5-1.x86_64.rpm
-sudo rpm -ivh influxdb-latest-1.x86_64.rpm
+sudo yum localinstall influxdb-latest-1.x86_64.rpm
 
 # for 32-bit systems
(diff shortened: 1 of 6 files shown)


beckettsean:
@otoolep @dgnorton @pauldix make sense?

beckettsean:
Thanks for the PR, @brianredbeard! 

Can you point me to the RedHat docs that describe this best practice? I'm fairly ignorant of the latest in distro-specific installation details.

otoolep:
Agreed. An explicit reference showing us that this is the right thing to do
would be great.

brianredbeard:
Absolutely:

Red Hat Documentation - [Working with Packages](https://access.redhat.com/documentation/en-US/Red_Hat_Enterprise_Linux/7/html/System_Administrators_Guide/sec-Working_with_Packages.html#sec-Installing)

You'll notice in that section that all packag management via RPM has been moved to an appendix as the direct use of `rpm` (much like the direct usage of `dpkg`) is discouraged.

otoolep:
OK, I checked the link you referenced. The change you proposed seems fair. 

Thanks @brianredbeard.

gunnaraasen:
Thanks @brianredbeard

PR #197, opened 2015-08-04
clarifying config file options

beckettsean:


beckettsean:
To address https://github.com/influxdb/influxdb/issues/3543#issuecomment-127460160

content/docs/v0.9/introduction/installation.md
```diff
@@ -58,12 +58,25 @@ For users who don't want to install any software and are ready to use InfluxDB,
...
+```shell
+/opt/influxdb/influxd -config /etc/influxdb/influxdb.generated.conf
 ```
 
+If no `-config` option is supplied, InfluxDB will use the default configuration file that ships with the installer. The actual location of the file depends on your OS. It is an identical file to the generated configuration file except that the default file has multiple comments.
+
+> Note: The `influxd` command has two similarly named flags. The `config` flag prints a generated default configuration file to STDOUT but does not launch the `influxd` process. The `-config` flag takes a single argument, which is the path to the InfluxDB configuration file to use when launching the process.
+
+
 ## Development Versions
```
mjdesa: If you start influxdb with `sudo service influxdb start` where does your config file need to be?

beckettsean: I don't know. I was hoping that @otoolep or someone would know.

content/docs/v0.9/introduction/installation.md
```diff
@@ -58,12 +58,25 @@ For users who don't want to install any software and are ready to use InfluxDB,
...
 ```
-/opt/influxdb/influxd config
+
+Edit the `influxdb.generated.conf` file to have the desired configuration settings. When launching InfluxDB, point the process to the correct configuration file using the `-config` option.
+
+```shell
+/opt/influxdb/influxd -config /etc/influxdb/influxdb.generated.conf
 ```
 
+If no `-config` option is supplied, InfluxDB will use the default configuration file that ships with the installer. The actual location of the file depends on your OS. It is an identical file to the generated configuration file except that the default file has multiple comments.
```
gunnaraasen: I think there's also another option, which is running `influxd` with no config provided at all. In that case, InfluxDB will use the built-in defaults, which should be the same as if the default values generated by `influxd -config`.

gunnaraasen: Actually, I tested this and my comment is not true. `influxd` gives an error when no `-config` option is given.

gunnaraasen: We should probably note that running `influxd` with no command is the same as running the `influxd run` command (which will return an error if the `-config .

PR #200, opened 2015-08-05
removing cardinality note

### LINKED ISSUE #184: remove the note about tag cardinality concerns
It's more of a red herring now than anything. With the series index not passing through raft consensus the cardinality of tags is much less relevant to performance. It certainly has an affect, but it's more of a 2nd order effect and thus not worth documenting until we have all the first order effects (CPU, RAM, iops, batch sizing, query tuning) hashed out.

beckettsean:
fixes #184


gunnaraasen:
+1 lets get rid of this.

PR #210, opened 2015-08-10
Various changes

gunnaraasen:
- Added an old blog post that was copied incorrectly from the old influxdb.org repo
- Fixed old images on the blog
- Removed all h1 headers from docs (all docs headers should be <h2)
- Normalized the style of notes to `> **Note:**`
- Cleaned up the query syntax page section on quoting identifiers and string literals


toddboom:
+1

beckettsean:
:+1:

PR #217, opened 2015-08-11
Replace smart quotes with normal quotes in code examples

Nibbler999:


beckettsean:
nice catch, thanks @Nibbler999

PR #218, opened 2015-08-11
Clustering

mjdesa:
This is a rough draft.


mjdesa:
@beckettsean @gunnaraasen looking for feedback.

In particular I feel like there is still a little bit of work that could be done in regards to fixing *common* clustering problems and diagnosing an unhealthy cluster. I've got a little bit in there, but it still feels lacking to me.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+
+## Configuration
+The following is the current recommended procedure for configuring a cluster. While it is still possible to configure your cluster using `peers` in the `[meta]` section of your config file, we encourage the use of the `-join` flag instead.
+
+### Start the Initial Raft Cluster
+
+Throughout this example, each node will be given a number that denotes the order in which it was started (e.g. 1 for the first node, 2 for the second node, etc).
+
+1. Install InfluxDB on the 3 machines.
+2. For each nodes `/etc/opt/influxdb/influxdb.conf` file, replace `hostname = "localhost"` with your hosts actual name.
```
beckettsean: node's (it's possessive)

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+3. For each nodes `/etc/opt/influxdb/influxdb.conf` file, update the `bind-address` to another port if `8088` is unacceptable.
```
beckettsean: node's

beckettsean: Do all servers have to use the same port, or can one use 8088, another use 9099, and the third use 10101?

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+4. Start InfluxDB on the first node.
+5. In `/etc/init.d/influxdb` on the second node, set `INFLUXD_OPTS="-join hostname_1:bind-address_1"`.
```
beckettsean: So we're using the environment variable to set a flag? Couldn't I also pass the `-join` flag directly at startup? (I get that the env variable way works on restart, but we should tell users about the `-join` method if it still works.)

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+
+| id | cluster_addr | raft |
+|----|--------------|------|
+|  1 | "hostname_1:bind-address_1" |  true |
+|  2 | "hostname_2:bind-address_2" |  true |
+|  3 | "hostname_3:bind-address_3" |  true |
+
+If you do not see all three raft nodes, your cluster is not healthy.
+
+> **Warning:** If you're having a hard time setting up your cluster, try settings the `/var/opt/influxdb/meta/peers.json` file manually to be `["<hostname 1>:<bind-address 1>","<hostname 2>:<bind-address 2>","<hostname 3>:<bind-address 1>"]` and `/var/opt/influxdb/meta/id` to be `1`, `2`, and `3` for each node respectively
```
beckettsean: setting, not settings

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Add More Data Nodes
+
+Once you have verified that your raft cluster is healthy and running appropriately, extra data nodes can be added.
+
+1. Install InfluxDB on the new node.
+2. In the new nodes `/etc/opt/influxdb/influxdb.conf` file, replace `hostname = "localhost"` with the nodes hosts actual name.
```
beckettsean: node's

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+3. In the new nodes `/etc/opt/influxdb/influxdb.conf` file, update the `bind-address` to another port if `8088` is unacceptable.
```
beckettsean: node's

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+2. In the new nodes `/etc/init.d/influxdb` file, set `INFLUXD_OPTS="-join hostname_1:bind-address_1,hostname_2:bind-address_2"`.
```
beckettsean: node's

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+3. Start InfluxDB on the new node.
+
+> **Note:** When using the `-join` you need only specify one `hostname:bind-address` pair. However, if more than one is provided, Influx will try to connect with the additional pairs in the case that it cannot connect with the first one.
```
beckettsean: Probably need a follow-up note here or a warning even, that this can lead to weird clusters of 2+2.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+| ...|        ...                  |  false |
+|  n | "hostname_n:bind-address_n" |  false |
+
+If you do not, then your node was not successfully added to the cluster.
+
+> **Warning** Currently InfluxDB supports writes to any node in the cluster, however queries must be directed at one of the 3 raft peers. This is a bug. See issue [3574](https://github.com/influxdb/influxdb/issues/3574) for more information.
+
+## Unimplemented Features
+
+* Add new raft members - If you have a one node cluster, and add a new node, it should eventually add the first 3 nodes as part of the raft cluster.  This is more involved and needs more work.  For now, all new nodes are data-only nodes.
```
beckettsean: I'd simplify "and add a new node" to "and add nodes"

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+title: Clustering
+aliases:
+  - /docs/v0.9/guides/clustering.html
+---
+
+> **Note:** Clustering is still in a beta state right now. There are still a good number of rough edges. If you notice any issues please [report them](https://github.com/influxdb/influxdb/issues/new).
+
+In 0.9.1 and 0.9.2 clusters are restricted 3 nodes and must be fully replicated, meaning all data is copied to all nodes and retention policies must have replication set to 3 for all three nodes in the cluster.
+
+Starting with version 0.9.3, Influxdb supports arbitrarily sized clusters that need not be fully replicated. Additionally new data nodes can be added to a cluster. The first three nodes to join a cluster are raft `peers`. All subsequent nodes are data nodes and do not participate in consensus. See Pull Request [#3478](https://github.com/influxdb/influxdb/pull/3478) for more information.
```
gunnaraasen: Do we need to reference the PR here? Seems like everything mentioned in the PR is addressed in this doc.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+|  n | "hostname_n:bind-address_n" |  false |
+
+If you do not, then your node was not successfully added to the cluster.
+
+> **Warning** Currently InfluxDB supports writes to any node in the cluster, however queries must be directed at one of the 3 raft peers. This is a bug. See issue [3574](https://github.com/influxdb/influxdb/issues/3574) for more information.
+
+## Unimplemented Features
+
+* Add new raft members - If you have a one node cluster, and add a new node, it should eventually add the first 3 nodes as part of the raft cluster.  This is more involved and needs more work.  For now, all new nodes are data-only nodes.
+* Join retries - If all the join addresses fail to join, the node will not retry need to be restarted to retry the join.
```
beckettsean: accidentally too many sentence here. Pick one.

content/docs/v0.9/guides/clustering.md
mjdesa: The init.d/influxdb script sets the environment variable. When you start influx as a service you cant pass in flags.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+* Leader RPC retries - remote calls to the raft cluster use a random member from the set of peers.  If that member is down, the call will fail and return an error.  Many of these could be retried automatically.
```
beckettsean: implications? how would I know? I'm not sure I get why I care about this.

content/docs/v0.9/guides/clustering.md
mjdesa: by cant I don't mean its impossible, it's just not possible without modifying the `init.d` script

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+---
+
+> **Note:** Clustering is still in a beta state right now. There are still a good number of rough edges. If you notice any issues please [report them](https://github.com/influxdb/influxdb/issues/new).
+
+In 0.9.1 and 0.9.2 clusters are restricted 3 nodes and must be fully replicated, meaning all data is copied to all nodes and retention policies must have replication set to 3 for all three nodes in the cluster.
+
+Starting with version 0.9.3, Influxdb supports arbitrarily sized clusters that need not be fully replicated. Additionally new data nodes can be added to a cluster. The first three nodes to join a cluster are raft `peers`. All subsequent nodes are data nodes and do not participate in consensus. See Pull Request [#3478](https://github.com/influxdb/influxdb/pull/3478) for more information.
+
+## Configuration
+The following is the current recommended procedure for configuring a cluster. While it is still possible to configure your cluster using `peers` in the `[meta]` section of your config file, we encourage the use of the `-join` flag instead.
```
gunnaraasen: > While it is still possible to configure your cluster using `peers` in the `[meta]` section of your config file, we encourage the use of the `-join` flag instead.

This should be a note that goes something like: "In versions 0.9.1 and 0.9.2, a cluster needs to be configured by specifying `peers`..."

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+
+### Start the Initial Raft Cluster
+
+Throughout this example, each node will be given a number that denotes the order in which it was started (e.g. 1 for the first node, 2 for the second node, etc).
+
+1. Install InfluxDB on the 3 machines.
```
gunnaraasen: Might be better to abstract this out into 3 InfluxDB instances so people who are testing a cluster locally or using Docker can follow along.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+2. For each node's `/etc/opt/influxdb/influxdb.conf` file, replace `hostname = "localhost"` with your hosts actual name.
```
gunnaraasen: "host's", not "hosts".

gunnaraasen: If I'm setting up a local cluster, can I use "localhost" for each instance?

content/docs/v0.9/guides/clustering.md
gunnaraasen: > This is more involved and needs more work.

Shouldn't this be "Configuring which nodes participate in raft consensus after the first three nodes form a cluster is not currently possible.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
+---
+title: Clustering
+aliases:
+  - /docs/v0.9/guides/clustering.html
```
gunnaraasen: This doc should continue aliasing `- /docs/v0.9/advanced_topics/clustering.html` and also be aliased to `- /docs/v0.9/concepts/clustering.html`, since page is being moved.

content/docs/v0.9/guides/clustering.md
mjdesa: I think the `-join` flag still works in 0.9.2, but I'm not completely certain.

gunnaraasen: Ok, we should note which release includes the `-join` functionality. The `[meta]` config options probably shouldn't be mentioned since it's not the recommended way to configure a cluster.

content/docs/v0.9/guides/clustering.md
mjdesa: I think that 2+2 clusters come up as the result of the initial cluster not being healthy, but I could be mistaken.

content/docs/v0.9/guides/clustering.md
mjdesa: Probably not. I don't have strong feelings about it. The language is slightly different in the PR and I though that maybe someone would find it helpful.

I'll rip it out unless there are any objections.

content/docs/v0.9/guides/clustering.md
mjdesa: I'm not sure how to set up a cluster locally because of port collision.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+
+At this point you'll want to verify that that your initial raft cluster is healthy. To do this, issue a `SHOW SERVERS` query to each node in your raft cluster. You should see something along the lines of this:
+
+| id | cluster_addr | raft |
+|----|--------------|------|
+|  1 | "hostname_1:bind-address_1" |  true |
+|  2 | "hostname_2:bind-address_2" |  true |
+|  3 | "hostname_3:bind-address_3" |  true |
+
+If you do not see all three raft nodes, your cluster is not healthy.
```
beckettsean: Maybe make some recommendations here as to what to try in order to make the cluster healthy? E.g. power down all nodes, restart in a certain order with `-join` flags, maybe change the `-join` order? Maybe they should delete the /data, /hh, and /raft directories before relaunching the cluster?

content/docs/v0.9/guides/clustering.md
jwilder: `[meta].peers` is being removed shortly.

content/docs/v0.9/guides/clustering.md
mjdesa: Do we have an advanced_topics sections?

content/docs/v0.9/guides/clustering.md
mjdesa: @beckettsean I'm actually not sure what to recommend. @jwilder Do you have any idea's of what would be good to try?

content/docs/v0.9/guides/clustering.md
jwilder: `[meta]` is now removed from master.  Should only use `-join` now.

content/docs/v0.9/guides/clustering.md
jwilder: You shouldn't set the id files unless you are trying to replace an existing node manually.  Those are assigned by the cluster when joining.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+> **Warning:** If you're having a hard time setting up your cluster, try settings the `/var/opt/influxdb/meta/peers.json` file manually to be `["<hostname 1>:<bind-address 1>","<hostname 2>:<bind-address 2>","<hostname 3>:<bind-address 1>"]` and `/var/opt/influxdb/meta/id` to be `1`, `2`, and `3` for each node respectively
+
+### Add More Data Nodes
+
+Once you have verified that your raft cluster is healthy and running appropriately, extra data nodes can be added.
+
+1. Install InfluxDB on the new node.
+2. In the new node's `/etc/opt/influxdb/influxdb.conf` file, replace `hostname = "localhost"` with the nodes hosts actual name.
+3. In the new node's `/etc/opt/influxdb/influxdb.conf` file, update the `bind-address` to another port if `8088` is unacceptable.
+2. In the new node's `/etc/init.d/influxdb` file, set `INFLUXD_OPTS="-join hostname_1:bind-address_1,hostname_2:bind-address_2"`.
```
jwilder: You can also pass `-hostname host[:port]` here instead of updating `hostname` in the config file.  This is the address that other nodes will use to connect to this node and should uniquely identify the host.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,72 @@
...
+3. Start InfluxDB on the new node.
+
+> **Note:** When using the `-join` you need only specify one `hostname:bind-address` pair. However, if more than one is provided, Influx will try to connect with the additional pairs in the case that it cannot connect with the first one.
```
jwilder: `hostname:port`

content/docs/v0.9/guides/clustering.md
jwilder: Adding new raft members works. The first 3 become raft peers and subsequent ones become data nodes.

content/docs/v0.9/guides/clustering.md
jwilder: I would remove this.  More of an implementation detail TODO.  Some places have retries now.

content/docs/v0.9/guides/clustering.md
jwilder: This has be fixed.  It will keep retrying each listed join addr every second until it succeeds.  This will run indefinitely until it either joins or the server is shutdown.

content/docs/v0.9/guides/clustering.md
beckettsean: Let's make it simple and say that anyone running clustering should be on 0.9.3 or later, or they are doing it wrong. 0.9.2 wasn't exactly a production-ready release

content/docs/v0.9/guides/clustering.md
beckettsean: not anymore but we used to, and this page was in there.

content/docs/v0.9/guides/clustering.md
beckettsean: the `hostname:bind-address` syntax came from the initial 0.9.1 clustering note that I think Philip wrote. To me it's less clear than `hostname:port`, but I don't know if it's somehow more correct.

@otoolep any objection to standardizing on `hostname:port` for the docs, instead of `hostname:bind-address`?

mjdesa: I chose `hostname:bind-address` because those are the names that get used in the `[meta]` section of the config file.

otoolep: No objection from me.

jwilder: Bind-addresses is a different concept and doesn't make sense in this context.

otoolep: Yeah, I agree with @jwilder

mjdesa:
@gunnaraasen @beckettsean @jwilder made changes everywhere that was suggested. Another quick look through would be appreciated.

content/docs/v0.9/guides/clustering.md
gunnaraasen: The alias creates a redirect for anyone who linked to the old page.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,79 @@
...
+The following is the current recommended procedure for configuring a cluster.
+
+> **Note:** In versions 0.9.1 and 0.9.2, a cluster needs to be configured by specifying peers. If you plan on using clustering it is highly recommended that you upgrade to version 0.9.3+.
+
+### Start the Initial Raft Cluster
+
+Throughout this example, each node will be given a number that denotes the order in which it was started (e.g. 1 for the first node, 2 for the second node, etc). It is also assumed that you are running some version of Linux and while it is possible to build a cluster locally, it is not recommended.
+
+1. Install InfluxDB on the 3 machines.
+2. For each node's `/etc/opt/influxdb/influxdb.conf` file, replace `hostname = "localhost"` with your host's actual name.
```
jwilder: I would add that this hostname must be resolvable by all members in the cluster.   It can be an IP or hostname and optional port number if necessary.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,79 @@
...
+3. For each node's `/etc/opt/influxdb/influxdb.conf` file, update the `bind-address` to another port if `8088` is unacceptable.
```
jwilder: The `bind-address` can also specify the host interface IP to use (e.g. `10.0.1.10:8088`).  By default, it will bind on all interfaces.

content/docs/v0.9/guides/clustering.md
```diff
@@ -0,0 +1,79 @@
...
+4. Start InfluxDB on the first node. Note that the `bind-address` may differ from node to node (e.g. one can use 8088, another use 9099, and the other 10101).
+5. In `/etc/init.d/influxdb` on the second node, set `INFLUXD_OPTS="-join hostname_1:bind-address_1"`.
```
jwilder: This should be the `hostname_1:port`, not the `hostname_1:bind-address_1`.  The `bind-address` can include an IP and be something like `10.0.1.8:8088`.  Other nodes should address that node as `host:8088`.  What is shown here would mean `host:10.9.1.8:8088` which is not valid.

mjdesa:
@jwilder fixed what you suggested. Let me know if you catch anything else.

beckettsean:
merging so we can push to new.influxdb.com

PR #219, opened 2015-08-13
Only values can have problems when a trying to insert a different type

sanrodari:
Only values can have problems when a trying to insert a different type. Tags don't have this problem.


gunnaraasen:
Thanks @sanrodari!

PR #224, opened 2015-08-14
clarify escaping

### LINKED ISSUE #173: Improve line protocol documentation
As noted in https://github.com/influxdb/influxdb/issues/3498

>It would be nice to improve a bit the line protocol documentation by 
 - adding = and \ to the list of chars to be escaped (maybe evident, but you never know) and 
 - re-specify the default unicode encoding used (utf8?) if applicable (or raw bytes if not).

>thanks

beckettsean:
partially fixes #173 (no note about default encoding)


gunnaraasen:
LGTM

Issue #116, opened 2015-07-14
new page: Document differences between 0.8 and 0.9


beckettsean:
- measurement + tagset in 0.9 is like series name in 0.8, so metadata moved from identifier to multiple identifiers.
- data store is now always BoltDB
- line protocol
- API endpoints simplified
- CLI
- admin GUI simplified in favor of HTTP API administration
- clustering redone entirely (CP + AP for metastore + data)
- InfluxQL re-written
- performance different
- CQs re-implemented

beckettsean:
closed by #225

PR #225, opened 2015-08-14
WIP: create 0.8 vs 0.9 page

beckettsean:


beckettsean:
fixes #116

content/docs/v0.9/concepts/08_vs_09
```diff
@@ -0,0 +1,78 @@
...
+
+The real change is evident when querying. In InfluxDB 0.8, querying multiple series required using a regular expression. For example, to query the `cpu` measurement where the `az` is `us-west-1` for all hostnames, the syntax is `SELECT * FROM /az.us-west-1.host..*.cpu$/ ...`. To query all series in the `cpu` measurement the syntax is `SELECT * FROM /cpu$/ ...`. 
+
+In InfluxDB 0.9, queries operate directly on measurements, not on series. All series in a measurement are returned by default, unless filtered out by tags in the WHERE clause. For example, to query the `cpu` measurement where the `az` is `us-west-1` for all hostnames, the syntax is `SELECT * FROM cpu WHERE az='us-west-1' ...`. To query all series in the `cpu` measurement the syntax is `SELECT * FROM cpu ...`.
```
pauldix: probably worth mentioning `GROUP BY <tag key>` in here somewhere. In 0.8 you could group by column names so that will feel similar. Also worth mentioning is that in 0.8 using `WHERE some_col = 'some_val'` was very inefficient. But with 0.9 doing `WHERE some_tag = 'some_val'` is very efficient and the preferred way to do things.

content/docs/v0.9/concepts/08_vs_09
```diff
@@ -0,0 +1,78 @@
...
+
+## Admin UI
+
+In InfluxDB 0.8 the Administration User Interface supported a number of configuration and adminsitration features, some of which could not be performed any other way. In InfluxDB 0.9 all database administration, configuration, and information is available via the HTTP API and the Admin UI has been substantially simplified. 
+
+## InfluxQL re-written
+
+As a result of the new schema, the query language was substantially overhauled. A number of functions available in InfluxDB 0.8 have not yet been re-implmented in InfluxDB 0.9, including [MODE](https://github.com/influxdb/influxdb/issues/1823), [HISTOGRAM](https://github.com/influxdb/influxdb/issues/3674), [DIFFERENCE](https://github.com/influxdb/influxdb/issues/1825), [TOP](https://github.com/influxdb/influxdb/issues/1821), [BOTTOM](https://github.com/influxdb/influxdb/issues/1820), and the [ORDER BY](https://github.com/influxdb/influxdb/issues/2022) clause. Links go to the open issues to re-implement each function or behavior.
+
+TODO: other justifications/improvements?
```
pauldix: Might be worth mentioning that it is now a pure Go binary, which makes it much easier to build for other architectures, like ARM. The query language in 0.8 was implemented with a C version of Yacc and Bison and the storage engine was in C++.

content/docs/v0.9/concepts/08_vs_09
```diff
@@ -0,0 +1,78 @@
...
+
+## Clustering 
+
+InfluxDB 0.9 follows a hybrid model for maintaining availibility of the data while guaranteeing consistency of the metadata. See CEO and co-founder Paul Dix's [clustering blog post](https://influxdb.com/blog/2015/06/03/InfluxDB_clustering_design.html) for more details.
+
+## Data Storage Engine
+
+In InfluxDB 0.8 users could choose from a selection of underlying storage engines, with RocksDB, LevelDB, HyperLevelDB, and LMDB as available options. In InfluxDB 0.9 the underlying storage engine is always [BoltDB](https://github.com/boltdb/bolt), a copy-on-write B+tree. 
+
+TODO: justification
```
pauldix: I'm going to write a blog post on this in the coming weeks, but in the meantime here's a quick word on it. There ended up being two primary reasons for this.

First, we wanted to have a pure Go binary to simplify our build chain and to make it easier to build for other OSes and platforms like Windows and ARM.

Second, to enable efficient eviction of old data through retention policies, we need to create an underlying database for each window of time. This ended up causing problems with the LSM Tree databases (LevelDB, RocksDB, and HyperLevelDB) because of the number of open file handles. People with large databases would have their processes crash because of too many file handles open.

The file handle problem also gave many users trouble because they wouldn't set their limits. This was the largest source of bug reports on the 0.8 line of versions. Users should still up their file handle limits for Influx, however they're now unlikely to hit that problem during their initial explorations with it.

beckettsean:
Updated with Paul's suggestions, merging so we can push to new.influxdb.com

PR #240, opened 2015-08-21
Clarify nightly build timezone

otoolep:


otoolep:
@beckettsean @gunnaraasen

content/site/download.md
```diff
@@ -46,7 +46,7 @@ Release candidates are made available for testing one week prior to release of t
         sudo rpm -ivh influxdb-0.9.3_rc1-1.x86_64.rpm
 
 ## Version 0.9.4 (Nightly)
-Nightly builds are created once-a-day using the top-of-tree of [master](https://github.com/influxdb/influxdb/tree/master) source code. These builds will include all the latest fixes, but also undergo only basic testing.
+Nightly builds are created once-a-day, at midnight Los Angeles, CA time, using the top-of-tree of [master](https://github.com/influxdb/influxdb/tree/master) source code. These builds will include all the latest fixes, but also undergo only basic testing.
```
beckettsean: Maybe say "at midnight Pacific Time"

otoolep: I thought about that, and originally wrote "Pacific Time", but I think that is a US-only term and don't want to be US-centric. But if it's a common term understood worldwide, cool (I think it probably is fine).

beckettsean: I can't find a definitive list of Time Zone names. Ultimately everyone wusses out and falls back to UTC-4 or whatever. Let's leave it as is. (We can probably remove the CA if we want to, since "Los Angeles" stands on its own without clarification according to the AP Style guide.)

beckettsean:
:+1: with or without modification to the name of the build time zone.

content/site/download.md
otoolep: +1 on removing CA if you like.

PR #241, opened 2015-08-21
Update download.md with correct links

kubum:
Apparently the latest stable is 0.9.2.1 rather than 0.9.2. It's a bit confusing for people how are visiting website without checking the github.

Is it a missed point or there is a reason to have a link to 0.9.2?


beckettsean:
0.9.2.1 was mostly an internal release to support some challenges with the hosted product. Thanks for updating the page

PR #252, opened 2015-08-25
make content and structural changes to querying edits guide

rkuchan:
Made changes to the structure of the Querying Data guide and added a section on the `chunk_size` parameter.

### DIFF
--- a/content/docs/v0.9/guides/querying_data.md
+++ b/content/docs/v0.9/guides/querying_data.md
@@ -4,13 +4,15 @@ alias:
   -/docs/v0.9/query_language/querying_data.html
 ---
 
-The HTTP API is also the primary means for querying data contained within InfluxDB. To perform a query send a `GET` to the endpoint `/query`, set the URL parameter `db` as the target database, and set the URL parameter `q` as your query. An example query, sent to a locally-running InfluxDB server, is shown below.
+The HTTP API is the primary means for querying data in InfluxDB. To perform a query send a `GET` to the endpoint `/query`, set the URL parameter `db` as the target database, and set the URL parameter `q` as your query. 
+
+The example below uses the HTTP API to query the same database that you encountered in [Writing Data](../guides/writing_data.html).
 
 ```sh
-curl -G 'http://localhost:8086/query' --data-urlencode "db=mydb" --data-urlencode "q=SELECT value FROM cpu_load_short WHERE region='us-west'"
+curl -G 'http://localhost:8086/query?pretty=true' --data-urlencode "db=mydb" --data-urlencode "q=SELECT value FROM cpu_load_short WHERE region='us-west'"
 ```
 
-Which returns:
+InfluxDB returns JSON. The results of your query appear in the `"results"` array. If an error occurs, InfluxDB sets an `"error"` key with an explanation of the error.
 
 ```json
 {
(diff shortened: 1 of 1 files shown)


content/docs/v0.9/guides/querying_data.md
```diff
@@ -4,13 +4,15 @@ alias:
...
 ---
 
-The HTTP API is also the primary means for querying data contained within InfluxDB. To perform a query send a `GET` to the endpoint `/query`, set the URL parameter `db` as the target database, and set the URL parameter `q` as your query. An example query, sent to a locally-running InfluxDB server, is shown below.
+The HTTP API is the primary means for querying data in InfluxDB. To perform a query send a `GET` to the endpoint `/query`, set the URL parameter `db` as the target database, and set the URL parameter `q` as your query. 
+
+The example below uses the HTTP API to query the same database that you encountered in [Writing Data](../guides/writing_data.html).
 
 ```sh
-curl -G 'http://localhost:8086/query' --data-urlencode "db=mydb" --data-urlencode "q=SELECT value FROM cpu_load_short WHERE region='us-west'"
+curl -G 'http://localhost:8086/query?pretty=true' --data-urlencode "db=mydb" --data-urlencode "q=SELECT value FROM cpu_load_short WHERE region='us-west'"
```
beckettsean: nice catch!

content/docs/v0.9/guides/querying_data.md
```diff
@@ -40,70 +46,86 @@ Which returns:
...
-```
-
-## Authentication
-Authentication is disabled by default. If authentication is enabled, user credentials must be supplied with every query. These can be suppled via the URL parameters `u` and `p`. For example, if the user is "bob" and the password is "mypass", then endpoint URL should take the form `/query?u=bob&p=mypass`.
+#### Authentication
+Authentication is disabled by default. If [authentication is enabled](../administration/config.html#authentication) you must supply user credentials with every query. You can supply them with the URL parameters `u` and `p`. For example, if `marty` is the user and the password is `hoverboard` then the endpoint URL should take the form `/query?u=marty&p=hoverboard`.
 
-The credentials may also be passed using _Basic Authentication_. If both types of authentication are present in a request, the URL parameters take precedence.
+Credentials may also be passed using [Basic Auth](../administration/authentication.html). Note that if both types of authentication are present in a request the URL parameters take precedence.
```
beckettsean: I would not link "Basic Auth" to anything, or if you want to, link it to http://tools.ietf.org/html/rfc2617, which is the actual spec. There's not really anything relevant on the linked docs page except that URL.

content/docs/v0.9/guides/querying_data.md
```diff
@@ -40,70 +46,86 @@ Which returns:
...
 
-## Pretty Printing
-When working directly with the API it is often convenient to have pretty-printed JSON output. To enable pretty-printed output, append `pretty=true` to the URL. For example:
+#### Chunk size
+For large queries, results are returned in batches of 10,000 points unless you use the query string parameter `chunk_size` to explicitly set the batch size. For example, get your results in batches of 20,000 points with:
 
 ```sh
-curl -G 'http://localhost:8086/query?pretty=true' --data-urlencode "db=mydb" --data-urlencode "q=SELECT * FROM cpu_load_short"
+curl -G 'http://localhost:8086/query' --data-urlencode "db=deluge" --data-urlencode "chunk_size=20000" --data-urlencode "q=SELECT * FROM liters"
```
beckettsean: cute, I like it.

beckettsean:
One link I would undo, otherwise :+1: :100:

content/docs/v0.9/guides/querying_data.md
gunnaraasen: I personally find the [Wikipedia article](https://en.wikipedia.org/wiki/Basic_access_authentication) to be more digestible than the spec.

It may be best to just remove most of this section, state that authentication is disabled by default, and point them to the authentication page for more info.

beckettsean: That's probably best. Fewer places to keep in sync as auth changes, and it's mostly extraneous to the goal of this page.

rkuchan:
Incorporated the edits. The biggest changes are in the authentication section and I did some reformatting.

beckettsean:
:+1:

Issue #258, opened 2015-08-26
field key in glossary is confusing


beckettsean:
We mention how to write integers, but not strings (must double-quote). I don't think we should mention syntax in the glossary unless absolutely necessary, that's better covered in the "Data Types" section in https://influxdb.com/docs/v0.9/write_protocols/write_syntax.html#line-protocol

mjdesa:
@beckettsean so if I understand correctly, you would like me to remove anything syntax related from the glossary to the `write_syntax`?

beckettsean:
Pretty much. Mostly it's weird that we mention `i` for ints, but not `"` for strings, for instance. Overall, though, as a glossary, it should describe the meanings, not the syntax for use.

mjdesa:
sounds good

mjdesa:
closed by https://github.com/influxdb/influxdb.com/pull/264

PR #264, opened 2015-08-27
Remove integer example from glossary

mjdesa:
In response to https://github.com/influxdb/influxdb.com/issues/258 the
mention of integers was removed from the `glossary`


beckettsean:
:+1:

PR #301, opened 2015-09-10
Make a couple small edits to the writing data guide

rkuchan:
Added a note about the line protocol in the first example and fixed the explanation of the 5xx HTTP response.


beckettsean:
:+1:

Issue #170, opened 2015-07-29
Document SHOW SERVERS


beckettsean:
Cluster admin page.

gunnaraasen:
Does `SHOW SERVERS` work right now?

beckettsean:
it does for me in 0.9.2

```
> show servers
id	url
1	http://127.0.0.1:8088

```

beckettsean:
tackle this as part of #171, probably

beckettsean:
@mjdesa can you add an example of actual `SHOW SERVERS` output on the clustering page? Once that's done we can close this.

PR #304, opened 2015-09-10
Add note about the output of show servers from cli

mjdesa:
Feels contrived, but fixes https://github.com/influxdb/influxdb.com/issues/170

### DIFF
--- a/content/docs/v0.9/guides/clustering.md
+++ b/content/docs/v0.9/guides/clustering.md
@@ -44,6 +44,16 @@ At this point you'll want to verify that that your initial raft cluster is healt
 |  2 | "hostname_2:port_2" |  true |
 |  3 | "hostname_3:port_3" |  true |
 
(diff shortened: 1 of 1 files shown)


beckettsean:
Do we need the preceding table or can we replace it with this new CLI output?

mjdesa:
Could be just me, but I think that the table looks a lot better than the CLI output.

9sx1w3vlhjdkf85nmpbbuvvjcjei2p6wtwbibe241yi

beckettsean:
Cool, works for me. :+1:

mjdesa:
Just a clarification. Should I keep both, or just the table or just the cli output

beckettsean:
You, @rkuchan, and @gunnaraasen should decide. I'm good with whatever the consensus is.

rkuchan:
I like the CLI output - if the columns can line up.

PR #316, opened 2015-09-15
Add post for 0.9.4 announcement

pauldix:


content/blog/2015/09/16/InfluxDB-0_9_4-released-with-top-order-by-time-desc-and-more.md
```diff
@@ -0,0 +1,40 @@
...
+
+```sql
+-- basic stats information
+show stats
+
+-- diagnostics information
+show diagnostics
+```
+
+We don't have statistics instrumentation for every part of the system yet, but we'll adding to it over time. Bt default, the server will write the stats information into a database called `_internal`. This data is only kept around for 7 days, but it'll be a first stop when we're helping people troubleshoot problems with InfluxDB.
```
otoolep: This is not quite right. What happens is that users *can* select the retention policy through the TOML file if they wish. If they do not explicitly choose the policy, data will be written to the autogenerated "default" policy, which has infinite retention. They can override this, but then they must create and configure the policy directly. 

https://github.com/influxdb/influxdb/blob/master/etc/config.sample.toml#L94

This was done to avoid the problem of the system creating and configuring a non-default policy for internal monitoring, the user then changing the details of that policy, and the system then changing it back again on a restart, thereby losing their changes.

otoolep: Actually, it's worse than that. We we reverted to this design, I nuked the policy control option in error.

I can add it back in again, but it will be in a nightly. Better remove any reference to the policy right now.

pauldix: The default RP for this data should not be infinite. Not if we're going to save it by default. If we're going to save it by default, it should be limited to 7 days. Otherwise, we should only save the data if the user has actually asked for it to be saved via the config.

pauldix: then we should cut a new 0.9.4.1 because keeping the data forever by default is not good.

otoolep:
Retention policy changes committed to master and merged to 0.9.4.

otoolep:
I see this has merged, but with all the brouhaha today over the move to Go 1.5, perhaps it would be worth mentioning this in the blog post.

content/blog/2015/09/16/InfluxDB-0_9_4-released-with-top-order-by-time-desc-and-more.md
otoolep: "Bt default..." -> "By default..."

toddboom:
@otoolep Just to mention that it's happened?

content/blog/2015/09/16/InfluxDB-0_9_4-released-with-top-order-by-time-desc-and-more.md
toddboom: Caught that one too. Thanks!

otoolep:
It's up to @pauldix , of course, but since one must actually build the surce with 1.5 now, it's seems worth calling out. And GC behaviour is going to be very different in 0.9.4, there should be no long pauses.

toddboom:
@otoolep ok, added a note about Go 1.5

Issue #153, opened 2015-07-23
[new page] create page for config file settings


beckettsean:
Document all config file options on a single page

beckettsean:
See https://influxdb.com/docs/v0.8/advanced_topics/configuration_options.html for inspiration, but let's add more detail than the comments from config.sample

beckettsean:
closed by #325

PR #325, opened 2015-09-17
restructure the config doc and add material [WIP]

rkuchan:
This is a first attempt at https://github.com/influxdb/influxdb.com/issues/153. 

There are a couple things to note/please comment on:

* Is the current layout OK or is the structure of the [v0.8 config doc](https://influxdb.com/docs/v0.8/advanced_topics/configuration_options.html) preferable?
* 🎁s mark where I wasn't sure what to write - it covers areas where I don't know what the options do as well as areas where it might be good to put a note about what works for most systems (but I wasn't sure what that was). Any knowledge gifts are appreciated!
* I linked to whatever documentation we have for the plugins but, in general, I left those sections well alone.
* I made all of my changes to administration/config.md - happy to make this a whole new page if that's preferred.

### DIFF
--- a/content/docs/v0.9/administration/config.md
+++ b/content/docs/v0.9/administration/config.md
@@ -155,7 +155,7 @@ This section controls the hinted handoff feature, which allows nodes to temporar
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `false` to disable hinted handoff. 
 
 **dir = "/var/opt/influxdb/hh"**  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The hinted handoff directory. For best throughput, the HH directory and the WAL directory should be on different physical devices. If you have performance concerns, you will also want to make this setting different from the dir in the [[data]](../administration/config.html#data) section.
 
 **max-size = 1073741824**  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum size of the hinted handoff queue for a node. If the queue is full, new writes are rejected and an error is returned to the client. The queue is drained when either the writes are retried successfully or the writes expire.  
@@ -213,7 +213,7 @@ This section controls InfluxDB's [system self-monitoring](https://github.com/inf
 By default, InfluxDB writes the data to the `_internal` database. If that database does not exist, InfluxDB creates it automatically. The `DEFAULT` retention policy on the `_internal` database is seven days. If you want to use a retention policy other than the seven-day retention policy, you must [create](../administration/administration.html#retention-policy-management) it. 
 
 **store-enabled = true**   
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `false` to disable recording statistics internally.
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `false` to disable recording statistics internally. If set to `false` it will make it substantially more difficult to diagnose issues with your installation.  
 
 **store-database = "_internal"**  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The destination database for recorded statistics.
@@ -248,13 +248,13 @@ This section controls how InfluxDB configures the HTTP endpoints. These are the 
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The port used by the HTTP API.
 
 **auth-enabled = false**  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `true` to enable authentication.
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `true` to require authentication.
 
 **log-enabled = true**  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `false` to disable logging.
 
 **write-tracing = false**  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `true` to enable logging for the write payload. 
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `true` to enable logging for the write payload. If set to `true`, this will duplicate every write statement in the logs and is thus not recommended for general use.  
 
 **pprof-enabled = false**  
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set to `true` to enable [pprof](http://blog.golang.org/profiling-go-programs) on InfluxDB so that it gathers detailed performance information. 
@@ -415,14 +415,15 @@ This section controls how [continuous queries (CQs)](../concepts/glossary.html#c
 &nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The upper bound on the number of previous interval queries that InfluxDB executes per CQ batch.
 
 **recompute-no-older-than = "10m0s"**  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;InfluxDB does not resample points with timestamps that are less than `now()` - `recompute-no-older-than`.
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;InfluxDB will not generate queries with an upper time boundary older than `now()` - `recompute-no-older-than`, regardless of the value of `recompute-previous-n`.
 
 **compute-runs-per-interval = 10**  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum number of generated queries per `GROUP BY time()` interval. The number of generated queries can be lower; this depends on the `GROUP BY time()` interval in the CQ and `compute-no-more-than`.
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The upper bound on the number of incremental queries generated within each `GROUP BY time()` interval. The actual number of generated queries can be lower, depending on the `GROUP BY time()` interval in the CQ and the `compute-no-more-than` setting.
 
 **compute-no-more-than = "2m0s"**  
-&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum frequency of CQ batch execution. This varies per CQ.
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Batches of CQs run at intervals determined by the `GROUP BY time()` interval divided by `compute-runs-per-interval`. However, CQ batches will never run more often than the `compute-no-more-than` value. 
 
+> **Note:** `GROUP BY time()` * (`recompute-previous-n` + 1) must be greater than `compute-no-more-than` or some time intervals will never be sampled.
 
 
 
--- a/content/docs/v0.9/administration/config.md
+++ b/content/docs/v0.9/administration/config.md
@@ -2,48 +2,372 @@
 title: Database Configuration
 ---
 
-## Generating a Configuration file
-
-To generate an InfluxDB configuration file run the command
-
+## The basics
+---
+Generate a new configuration file by running `influxd config` and redirecting the output to a file:
 ```
-influxd config  > influx.conf
-```
-
-## Authentication
-
-To add authentication to InfluxDB set `auth-enabled = true` in the `[http]` section of your config file.
-
-```
-[http]
-...
-auth-enabled = true
-...
+influxd config  > /etc/influxdb/influx_v0.9.x.conf
 ```
 
-## Anonymous Statistics
-
-By default, InfluxDB sends anonymous statistics about your InfluxDB instance. If you would like to disable this functionality, set `enabled = false` in the `[monitoring]` section of your config file.
-
+Start the influxd process using that configuration file with `influxd -config`:  
 ```
-[monitoring]
-enabled = false
+influxd -config /etc/influxdb/influx_v0.9.x.conf
 ```
 
-## Cluster
+## Upgrading your configuration file
+___
+While configuration files from prior versions of InfluxDB 0.9 should work with future releases, old files may lack options for new features. To keep your configuration file up-to-date we recommend doing one of the following:
 
-_Note: Clustering is in alpha state right now and all clusters __must__ contain three and only three nodes._
+* Generate a new configuration file with each InfluxDB upgrade. Any changes that you made to the old file will need to be manually ported to the new file.
 
-Configuring a cluster with host machines A, B, and C:
+* Use both the `config` and `-config` flags to combine the new configuration file with your old configuration file. In the example below, InfluxDB uses the union of the new configuration file and the old configuration file (`influx_v0.9.x.conf`), where any configurations in the old file trump those in the new file.  
+<br>
+    ```
+    influxd config -config /etc/influxdb/influx_v0.9.x.conf
+    ```
 
-1. Install InfluxDB on all three machines.
-2. Generate a config file on each machine by running `influxd config > influxdb.conf`.
-3. Update the `[meta]` section of the configuration file on all three hosts, replacing `localhost` with the hosts actual network IP.
-4. Update the bind-address to another port if 8088 is not acceptable.
-5. On all three machines add `peers = ["A_IP:A_bindaddress", "B_IP:B_bindaddress", "C_IP:C_bindaddress"]` to the `[meta]` section of the config file.
-6. On all three machines, add `replication = 3` in the `[retention]` section of the config file.
-7. Launch `influxd` on hosts A, B, and C in order.
+## The configuration file
+---
+The following sections follow the structure of the [sample configuration file on GitHub](https://github.com/influxdb/influxdb/blob/master/etc/config.sample.toml) and offer detailed explanations of the different options.
 
-## A Note about `dir` in `[meta]` and `[data]`
+Configuration sections:  
 
-In both the `[meta]` and `[data]` the `dir` configuration setting must be on the same filesystem.
+* [reporting](../administration/config.html#reporting)  
+* [[meta]](../administration/config.html#meta)  
+* [[data]](../administration/config.html#data)  
+* [[cluster]](../administration/config.html#cluster)  
+* [[retention]](../administration/config.html#retention)  
+* [[monitor]](../administration/config.html#monitor)  
+* [[admin]](../administration/config.html#admin)  
+* [[http]](../administration/config.html#http)  
+* [[graphite]](../administration/config.html#graphite)  
+* [[collectd]](../administration/config.html#collectd)  
+* [[opentsdb]](../administration/config.html#opentsdb)  
+* [[udp]](../administration/config.html#udp)  
+* [[continuous_queries]](../administration/config.html#continuous-queries)  
+* [[hinted-handoff]](../administration/config.html#hinted-handoff)  
+
+## reporting
+Once every 24 hours InfluxDB reports anonymous data to m.influxdb.com. Those data include Raft ID (random 8 bytes), OS, architecture, InfluxDB version, and metadata. InfluxDB doesn't track IP addresses of servers reporting. This is only used to track the number of instances running and the versions, which is very helpful for us. 
+
+**reporting-disabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Change this option to `true` to disable reporting.
+
+## [meta]
+This section handles some of the parameters for the InfluxDB cluster. Specifically, it controls the parameters for the Raft consensus group which stores metadata about the cluster. For step-by-step instructions on setting up a cluster, see the [setup guide](../guides/clustering.html). 
+
+**dir = "/var/opt/influxdb/meta"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The directory where InfluxDB stores `id`, `peers.json`, `raft.db`, and the `snapshots` directory.
+
+> * `id` stores the identification number of the Raft peer: `1` for the first node to join the cluster, `2` for the second node, and `3` for the third node to join the cluster.
+* `peers.json` stores the hostnames and ports of the three Raft peers.
+* `snapshots` contains the server's snapshots taken for the purpose of log compaction.
+* `raft.db` 🎁
+
+**hostname = "localhost"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The `hostname` of the Raft peer. 
+
+**bind-address = ":8088"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The `port` over which the Raft peer communicates.
+
+**retention-autocreate = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**election-timeout = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time that the follower waits for communication from the leader before the follower begins a new election. In practice, InfluxDB staggers this parameter for each 
+Raft peer. The default should work for most systems.
+
+**heartbeat-timeout = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The rate at which the leader sends heartbeats to the other Raft peers to preserve its leader status. You may want to alter this parameter depending on your network.
+
+**leader-lease-timeout = "500ms"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**commit-timeout = "50ms"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+## [data]
+This section controls where the actual shard data for InfluxDB live and how they are flushed from the WAL. You may want to change the `dir` setting, but the WAL settings are an advanced configuration. The defaults should work for most systems.
+
+**dir = "/var/opt/influxdb/data"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The directory where InfluxDB stores the shard data.
+
+The following three WAL settings are for the b1 storage engine used in 0.9.2. They won't apply to any new shards created after upgrading to versions 0.9.3+. 
+ 
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; **max-wal-size = 104857600**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum size the WAL can reach before a flush. Defaults to 100MB.
+  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-flush-interval = "10m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum time data can sit in WAL before a flush.
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-partition-flush-delay = "2s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The delay time between each WAL partition being flushed.
+
+The following WAL settings are for the storage engine in versions 0.9.3+.
+ 
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-dir = "/var/opt/influxdb/wal"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+ 
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-enable-logging = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging.
+
+🎁Any explanation for why the next lines are commented out?
+
+**# wal-ready-series-size = 25600**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;When a series in the WAL in-memory cache reaches this size in bytes it is marked as ready to flush to the index.
+  
+**# wal-compaction-threshold = 0.6**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Flush and compact a partition once this ratio of series are over the ready size.
+
+**# wal-max-series-size = 2097152**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a flush and compaction if any series in a partition gets above this size in bytes.
+
+**# wal-flush-cold-interval = "10m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a flush of all series and full compaction if there have been no writes in this amount of time. This is useful for ensuring that shards that are cold for writes don't keep a bunch of data cached in memory and in the WAL.
+
+**# wal-partition-size-threshold = 20971520**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a partition to flush its largest series if it reaches this approximate size in bytes. Remember there are 5 partitions so you'll need at least 5x this amount of memory. The more memory you have, the bigger this can be.
+
+## [cluster]
+This section handles non-Raft cluster behavior, which generally includes how data are shared across shards.
+
+**shard-writer-timeout = "5s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time within which a shard must respond to a write.
+ 
+ **write-timeout = "5s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time within which a write operation must complete on the cluster.
+
+## [retention]
+This section controls the enforcement of retention policies for evicting old data.
+
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Change this to `false` to prevent InfluxDB from enforcing retention policies. 
+
+**check-interval = "30m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The rate at which InfluxDB checks to enforce a retention policy. 
+
+## [monitor]
+This section controls InfluxDB's [system self-monitoring](https://github.com/influxdb/influxdb/blob/master/monitor/README.md).
+
+The retention policy for these data is the default retention policy within the `_internal` database. If the `_internal` database does not exist, InfluxDB creates that database automatically with a default retention policy of infinite retention. If you want to use a retention policy other than infinite retention, you must [create](../administration/administration.html#retention-policy-management) it. 
+
+**store-enabled = true**   
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Whether to record statistics internally.
+
+**store-database = "_internal"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The destination database for recorded statistics.
+
+**store-interval = "10s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The interval at which to record statistics.
+
+## [admin]
+Controls the availability of the built-in, web-based admin interface.
+
+>**Note:** If you want to enable HTTPS for the admin interface you must also enable HTTPS on the [[http]](../administration/config.html#http) service.
+
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable the admin interface.
+
+**bind-address = ":8083"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The admin interface uses port `8083` by default.
+
+**https-enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable HTTPS for the admin interface.
+
+**https-certificate = "/etc/ssl/influxdb.pem"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to the path of the certificate file.
+
+## [http]
+This section controls how InfluxDB configures the HTTP endpoints. These are the primary mechanism for getting data into and out of InfluxDB. Edit options in this section to enable HTTPS and authentication.
+
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable HTTP. Note that the InfluxDB CLI connects to the database using the HTTP API.
+
+**bind-address = ":8086"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The HTTP API uses port `8086`  by default.
+
+**auth-enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable authentication.
+
+**log-enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging.
+
+**write-tracing = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**pprof-enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**https-enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable HTTPS.
+
+**https-certificate = "/etc/ssl/influxdb.pem"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to the path of the certificate file.
+
+## [[graphite]]
+This section controls one or many listeners for Graphite data. See the [README](https://github.com/influxdb/influxdb/blob/master/services/graphite/README.md) on GitHub for more information.
+
+**enabled = false**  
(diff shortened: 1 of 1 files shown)


content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
-To add authentication to InfluxDB set `auth-enabled = true` in the `[http]` section of your config file.
+* Generate a new configuration file with each InfluxDB upgrade. Any changes that you made to the old file will need to be manually ported to the new file.
 
-```
-[http]
-...
-auth-enabled = true
-...
-```
+* Use both the `config` and `-config` flags to combine the new configuration file with your old configuration file. In the example below, InfluxDB uses the union of the new configuration file and the old configuration file (`influx_v0.9.x.conf`), where any configurations in the old file trump those in the new file.
```
beckettsean: "trump" does not translate well, perhaps "take precedence"

beckettsean: Maybe link to https://influxdb.com/docs/v0.9/introduction/installation.html#generate-a-configuration-file for more detail?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
-```
+* Use both the `config` and `-config` flags to combine the new configuration file with your old configuration file. In the example below, InfluxDB uses the union of the new configuration file and the old configuration file (`influx_v0.9.x.conf`), where any configurations in the old file trump those in the new file.  
+<br>
+    ```
+    influxd config -config /etc/influxdb/influx_v0.9.x.conf
+    ```
+
+## The configuration file
+---
+The following sections follow the structure of the [sample configuration file on GitHub](https://github.com/influxdb/influxdb/blob/master/etc/config.sample.toml) and offer detailed explanations of the different options.
```
beckettsean: Note that this doc may be slightly out of date. This is a link to Master, meaning it will always be ahead of the last official release, which is usually what we document.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+Configuration sections:  
+
+* [reporting](../administration/config.html#reporting)
```
beckettsean: Should this be double-bracketed?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+* [[http]](../administration/config.html#http)  
+* [[graphite]](../administration/config.html#graphite)  
+* [[collectd]](../administration/config.html#collectd)  
+* [[opentsdb]](../administration/config.html#opentsdb)  
+* [[udp]](../administration/config.html#udp)  
+* [[continuous_queries]](../administration/config.html#continuous-queries)  
+* [[hinted-handoff]](../administration/config.html#hinted-handoff)  
+
+## reporting
+Once every 24 hours InfluxDB reports anonymous data to m.influxdb.com. Those data include Raft ID (random 8 bytes), OS, architecture, InfluxDB version, and metadata. InfluxDB doesn't track IP addresses of servers reporting. This is only used to track the number of instances running and the versions, which is very helpful for us.
```
beckettsean: Those data include a unique randomly-generated cluster identifier (an 8-byte Raft ID), ...

beckettsean: Also need to clarify "metadata", that's way to vague for this. Users could interpret it to mean nearly anything and that's scary.

beckettsean: "InfluxDB doesn't request, track, or store the IP addresses of those servers that report. "

beckettsean: "InfluxDB uses these data primarily to track the number of deployed clusters for each version."

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+* [[hinted-handoff]](../administration/config.html#hinted-handoff)  
+
+## reporting
+Once every 24 hours InfluxDB reports anonymous data to m.influxdb.com. Those data include Raft ID (random 8 bytes), OS, architecture, InfluxDB version, and metadata. InfluxDB doesn't track IP addresses of servers reporting. This is only used to track the number of instances running and the versions, which is very helpful for us. 
+
+**reporting-disabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Change this option to `true` to disable reporting.
+
+## [meta]
+This section handles some of the parameters for the InfluxDB cluster. Specifically, it controls the parameters for the Raft consensus group which stores metadata about the cluster. For step-by-step instructions on setting up a cluster, see the [setup guide](../guides/clustering.html).
```
beckettsean: I would re-word "which stores metadata" to something more active like "which coordinates metadata"

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The rate at which the leader sends heartbeats to the other Raft peers to preserve its leader status. You may want to alter this parameter depending on your network.
+
+**leader-lease-timeout = "500ms"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**commit-timeout = "50ms"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+## [data]
+This section controls where the actual shard data for InfluxDB live and how they are flushed from the WAL. You may want to change the `dir` setting, but the WAL settings are an advanced configuration. The defaults should work for most systems.
```
beckettsean: I would say "where the shards are stored" or "where the data are stored" but not both. The shards hold the data. We don't store data about the shards here.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+**dir = "/var/opt/influxdb/data"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The directory where InfluxDB stores the shard data.
```
beckettsean: same note as above

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+The following three WAL settings are for the b1 storage engine used in 0.9.2. They won't apply to any new shards created after upgrading to versions 0.9.3+.
```
beckettsean: Maybe we need a universal call-out for minor version-specific issues.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**dir = "/var/opt/influxdb/data"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The directory where InfluxDB stores the shard data.
+
+The following three WAL settings are for the b1 storage engine used in 0.9.2. They won't apply to any new shards created after upgrading to versions 0.9.3+. 
+ 
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp; **max-wal-size = 104857600**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum size the WAL can reach before a flush. Defaults to 100MB.
+  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-flush-interval = "10m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum time data can sit in WAL before a flush.
```
beckettsean: sit in _the_ WAL

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-partition-flush-delay = "2s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The delay time between each WAL partition being flushed.
+
+The following WAL settings are for the storage engine in versions 0.9.3+.
+ 
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-dir = "/var/opt/influxdb/wal"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
```
beckettsean: this is configurable because for best throughput, the `wal` and `data` directories should be on different physical devices. Thus anyone with performance concerns will want to make this setting different from the one in https://github.com/influxdb/influxdb.com/pull/325/files#diff-cd984a3ae159299cda132f8704411875R92

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+ 
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;**wal-enable-logging = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging.
+
+🎁Any explanation for why the next lines are commented out?
```
beckettsean: I think @pauldix shrunk the number of WAL partitions from 5 to 1, which might make the individual settings moot. I know of no other reason to comment them out.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Flush and compact a partition once this ratio of series are over the ready size.
+
+**# wal-max-series-size = 2097152**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a flush and compaction if any series in a partition gets above this size in bytes.
+
+**# wal-flush-cold-interval = "10m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a flush of all series and full compaction if there have been no writes in this amount of time. This is useful for ensuring that shards that are cold for writes don't keep a bunch of data cached in memory and in the WAL.
+
+**# wal-partition-size-threshold = 20971520**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a partition to flush its largest series if it reaches this approximate size in bytes. Remember there are 5 partitions so you'll need at least 5x this amount of memory. The more memory you have, the bigger this can be.
```
beckettsean: Check with @pauldix, I'm pretty sure there are no longer 5 partitions. That might be an 0.9.3 exclusive.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+## [cluster]
+This section handles non-Raft cluster behavior, which generally includes how data are shared across shards.
+
+**shard-writer-timeout = "5s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time within which a shard must respond to a write.
```
beckettsean: what happens if this limit is exceeded? the write fails and goes to hinted handoff?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+ 
+ **write-timeout = "5s"**
```
beckettsean: if this fails is the write rejected? Is this related to the consistency setting or does this timeout happen before or after that consistency setting would apply?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Change this to `false` to prevent InfluxDB from enforcing retention policies. 
+
+**check-interval = "30m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The rate at which InfluxDB checks to enforce a retention policy. 
+
+## [monitor]
+This section controls InfluxDB's [system self-monitoring](https://github.com/influxdb/influxdb/blob/master/monitor/README.md).
+
+The retention policy for these data is the default retention policy within the `_internal` database. If the `_internal` database does not exist, InfluxDB creates that database automatically with a default retention policy of infinite retention. If you want to use a retention policy other than infinite retention, you must [create](../administration/administration.html#retention-policy-management) it.
```
beckettsean: I believe this is no longer infinite retention. @otoolep did the 7 day retention default make it into 0.9.4.1? Are there any official releases where the default RP for `_internal` is infinite?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+**auth-enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable authentication.
 
-## Anonymous Statistics
+**log-enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging.
 
-By default, InfluxDB sends anonymous statistics about your InfluxDB instance. If you would like to disable this functionality, set `enabled = false` in the `[monitoring]` section of your config file.
+**write-tracing = false**
```
beckettsean: I believe this controls whether we log every write event or not. @otoolep?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
 
-## Anonymous Statistics
+**log-enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging.
 
-By default, InfluxDB sends anonymous statistics about your InfluxDB instance. If you would like to disable this functionality, set `enabled = false` in the `[monitoring]` section of your config file.
+**write-tracing = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
 
+**pprof-enabled = false**
```
beckettsean: @otoolep?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable the admin interface.
+
+**bind-address = ":8083"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The admin interface uses port `8083` by default.
+
+**https-enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable HTTPS for the admin interface.
+
+**https-certificate = "/etc/ssl/influxdb.pem"**
```
beckettsean: what happens if this is a different PEM from the one in the [http] section? Can they be different?

@toddboom @gunnaraasen?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+## [[graphite]]
+This section controls one or many listeners for Graphite data. See the [README](https://github.com/influxdb/influxdb/blob/master/services/graphite/README.md) on GitHub for more information.
+
+**enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this to `true` to enable Graphite input.
+
+**# bind-address = ":2003"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+
+**# protocol = "tcp"**
```
beckettsean: what are the options? TCP and UDP? Anything else?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this to `true` to enable Graphite input.
+
+**# bind-address = ":2003"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+
+**# protocol = "tcp"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+
+**# consistency-level = "one"**
```
beckettsean: controls writes. This should link to something about consistency. All I find right now is the query string parameter definition on https://influxdb.com/docs/v0.9/write_protocols/write_syntax.html

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+The next three options control how batching works. You should have this enabled otherwise you could get dropped metrics or poor performance. Batching will buffer points in memory if you have many coming in.
+
+**# batch-size = 1000**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The input will flush if this many points get buffered.
+
+**# batch-pending = 5**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The number of batches that may be pending in memory.
+
+**# batch-timeout = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The input will flush at least this often even if we haven't reached the configured batch size.
```
beckettsean: ... even if it hasn't reached ...

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+**# name-schema = "type.host.measurement.device"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;This option configures tag names for parsing the metric name from graphite protocol; separated by `name-separator`. The "measurement" tag is special and the corresponding field will become the name of the metric. e.g. "type.host.measurement.device" will parse "server.localhost.cpu.cpu0" as:
```
beckettsean: configures tag keys

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**ignore-unnamed = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;If set to `true`, when the input metric name has more fields than `name-schema` specified, the extra fields will be ignored. Otherwise an error will be logged and the metric rejected.
+
+## [collectd]
+This section controls the listener for Collectd data.
+
+**enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable Collectd writes.
+
+**# bind-address = ""**
```
beckettsean: no default? seems like there should be one

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+## [collectd]
+This section controls the listener for Collectd data.
+
+**enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable Collectd writes.
+
+**# bind-address = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+
+**# database = ""**
```
beckettsean: default to collectd?

content/docs/v0.9/administration/config.md
otoolep: Yes, the 7-day retention period made it in.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable Collectd writes.
+
+**# bind-address = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+
+**# database = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The name of the database that you want to write to.
+
+**# typesdb = ""**
```
beckettsean: I have no idea what this does

content/docs/v0.9/administration/config.md
otoolep: Yeah, we're down to 1 partition in 0.9.4. This is not configurable.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**# batch-timeout = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The input will flush at least this often even if we haven't reached the configured batch-size.
+
+## [opentsdb]
+Controls the listener for OpenTSDB data. See the [README](https://github.com/influxdb/influxdb/blob/master/services/opentsdb/README.md) on GitHub for more information.
+
+**enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable openTSDB writes.
+
+**# bind-address = ""**
```
beckettsean: again it seems like these three settings should have reasonable defaults, not just empty string

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The number of batches that may be pending in memory.
+
+**# batch-timeout = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Will flush at least this often even if we haven't reached the configured batch-size.
+
+## [continuous_queries]
+This section controls how continuous queries run within InfluxDB.
+
+**log-enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging.
```
beckettsean: ... to disable logging for continuous query events

(as opposed to disabling other logging or logging entirely)

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+**compute-runs-per-interval = 10**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**compute-no-more-than = "2m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+## [hinted-handoff]
+This section controls the hinted handoff feature, which allows nodes to temporarily store queued data when one node of a cluster is down for a short period of time.
+
+🎁Any notes on what may need to be changed for most systems?
```
beckettsean: HH has no function in a single node cluster. We might mention that here.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
+
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable hinted handoff. 
+
+**dir = "/var/opt/influxdb/hh"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+
+**max-size = 1073741824**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum size of the hinted handoff queue.
```
beckettsean: what happens if this is exceeded? Are older writes pushed off the stack by newer ones? Do new writes for that node fail? Return failures to the client?

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,372 @@
...
 
-_Note: Clustering is in alpha state right now and all clusters __must__ contain three and only three nodes._
+**max-age = "168h"**
```
beckettsean: do points age-out after max-age? do entire files age out? Are they silently dropped? is anything logged or returned to STDERR?

corylanou:
This actually brings up another issue, which is some of our packages aren't documented well enough in the code.

For instance, all of this config information should be readily available via godoc as well. So for the opentsdb package, you could reference this:

https://godoc.org/github.com/influxdb/influxdb/services/opentsdb

We might want to open issues for these areas and label them as `help-wanted` as well.  I can help create these if you think they make sense.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+* [[http]](../administration/config.html#http)  
+* [[graphite]](../administration/config.html#graphite)  
+* [[collectd]](../administration/config.html#collectd)  
+* [[opentsdb]](../administration/config.html#opentsdb)  
+* [[udp]](../administration/config.html#udp)  
+* [[continuous_queries]](../administration/config.html#continuous-queries)  
+* [[hinted-handoff]](../administration/config.html#hinted-handoff)  
+
+## reporting
+Once every 24 hours InfluxDB reports anonymous data to m.influxdb.com. Those data include a unique, randomly-generated cluster identifier (an 8-byte Raft ID), OS, architecture, InfluxDB version, and metadata (🎁 help on clarifying metadata - users could interpret it to mean nearly anything). InfluxDB doesn't request, track, or store the IP addresses of those servers that report. InfluxDB uses these data primarily to track the number of deployed clusters for each version.
```
pauldix: metadata is: number of databases, number of measurements, number of unique series

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+## [meta]
+This section handles some of the parameters for the InfluxDB cluster. Specifically, it controls the parameters for the Raft consensus group which coordinates metadata about the cluster. For step-by-step instructions on setting up a cluster, see the [setup guide](../guides/clustering.html). 
+
+**dir = "/var/opt/influxdb/meta"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The directory where InfluxDB stores `id`, `peers.json`, `raft.db`, and the `snapshots` directory.
+
+> * `id` stores the identification number of the Raft peer: `1` for the first node to join the cluster, `2` for the second node, and `3` for the third node to join the cluster.
+* `peers.json` stores the hostnames and ports of the three Raft peers.
+* `snapshots` contains the server's snapshots taken for the purpose of log compaction.
+* `raft.db` 🎁
```
pauldix: is the BoltDB database that contains the Raft log and snapshots.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+
+**hostname = "localhost"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The `hostname` of the Raft peer. 
+
+**bind-address = ":8088"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The `port` over which the Raft peer communicates.
+
+**retention-autocreate = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
```
pauldix: InfluxDB will create shards prior to when they're needed in order to avoid contention when a new block of time begins. This should be on by default.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
-## Anonymous Statistics
+**auth-enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable authentication.
 
-By default, InfluxDB sends anonymous statistics about your InfluxDB instance. If you would like to disable this functionality, set `enabled = false` in the `[monitoring]` section of your config file.
+**log-enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging.
 
+**write-tracing = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁Does this control whether we log every write event or not?
```
pauldix: yes, and other detailed write information

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+
+**pprof-enabled = false**
```
pauldix: pprof is a Go tool for doing performance analysis of a running Go program. This will enable it on InfluxDB so detailed performance information can be gathered. An introduction to pprof is here: http://blog.golang.org/profiling-go-programs

content/docs/v0.9/administration/config.md
pauldix: just those two

content/docs/v0.9/administration/config.md
pauldix: having it empty means it'll bind to all IPs. Equivalent to `0.0.0.0`

content/docs/v0.9/administration/config.md
pauldix: this is in the `udp` section so I'd expect that. Either that or we should validate the config and error out if this plugin is enabled but the db isn't set.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+## [continuous_queries]
+This section controls how continuous queries run within InfluxDB.
+
+**log-enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging for continuous query events.
+
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable continuous queries.
+
+**recompute-previous-n = 2**
```
dgnorton: When continuous queries are run we'll automatically recompute previous intervals
in case lagged data came in. Set to zero if you never have lagged data.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable continuous queries.
+
+**recompute-previous-n = 2**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**recompute-no-older-than = "10m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum interval over which the continuous query will recompute results.
+
+**compute-runs-per-interval = 10**
```
dgnorton: compute-runs-per-interval will determine how many times the current and previous N intervals
will be computed. The group by time will be divided by this and it will get computed  this many times:
group by time seconds / runs per interval
This will give partial results for current group by intervals and will determine how long it will
be until lagged data is recomputed. For example, if this number is 10 and the group by time is 10m, it
will be a minute past the previous 10m bucket of time before lagged data is picked up.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+**recompute-previous-n = 2**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**recompute-no-older-than = "10m"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum interval over which the continuous query will recompute results.
+
+**compute-runs-per-interval = 10**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**compute-no-more-than = "2m"**
```
dgnorton: compute-no-more-than paired with the compute-runs-per-interval will determine the ceiling of how many times smaller
 `GROUP BY` times will be computed. For example, if you have compute-runs-per-interval set to `10` and this setting
 to `1m`. Then for a group by `time(1m)` will actually only get computed once per interval (and once per previous-n).
 If you have a group by `time(5m)` then you'll get five computes per interval. Any group by time window larger
 than `10m` will get computed 10 times for each interval.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable openTSDB writes.
+
+**# bind-address = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁Default?
+
+**# database = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The name of the database that you want to write to. If the database does not exist, it will be created automatically when the input is initialized.🎁Default?
+
+**# retention-policy = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁Default?
```
dgnorton: I think it should be `default`.

content/docs/v0.9/administration/config.md
dgnorton: @pauldix this one is for automatically creating a default RP when a database is created.  https://github.com/influxdb/influxdb/blob/master/meta/store.go#L872

If this is set to `true`, a retention policy named `default` is automatically created and set as the default policy when a database is created.  The replication factor is set to the number of nodes in the cluster (i.e., fully replicated).

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a flush of all series and full compaction if there have been no writes in this amount of time. This is useful for ensuring that shards that are cold for writes don't keep a bunch of data cached in memory and in the WAL.
+
+**# wal-partition-size-threshold = 20971520**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a partition to flush its largest series if it reaches this approximate size in bytes. Remember there are five partitions so you'll need at least five times this amount of memory. The more memory you have, the bigger this can be.
+
+## [cluster]
+This section handles non-Raft cluster behavior, which generally includes how data are shared across shards.
+
+**shard-writer-timeout = "5s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time within which a shard must respond to a write.🎁 What happens if this limit is exceeded? The write goes to hinted handoff?
```
jwilder: This is the time that a write from one node to another must complete before the write times out.  If it times out, it may still succeed on the remote node but the client node stops waiting and queues it in hinted handoff.  This value should always be less than the `write-timeout`.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable Collectd writes.
+
+**# bind-address = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁 No default?
+
+**# database = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The name of the database that you want to write to. 🎁 Default to collectd?
+
+**# typesdb = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
```
corylanou: Location of the local file that provides data-set specifications for the system statistics collection daemon collectd.  Defaults to `/usr/share/collectd/types.db`

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+This section controls the listener for Collectd data.
+
+**enabled = false**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `true` to enable Collectd writes.
+
+**# bind-address = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁 No default?
+
+**# database = ""**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The name of the database that you want to write to. 🎁 Default to collectd?
```
corylanou: Defaults to `collectd`: https://github.com/influxdb/influxdb/blob/master/services/collectd/config.go#L12

content/docs/v0.9/administration/config.md
corylanou: https://github.com/influxdb/influxdb/blob/master/services/collectd/config.go#L22

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Force a partition to flush its largest series if it reaches this approximate size in bytes. Remember there are five partitions so you'll need at least five times this amount of memory. The more memory you have, the bigger this can be.
+
+## [cluster]
+This section handles non-Raft cluster behavior, which generally includes how data are shared across shards.
+
+**shard-writer-timeout = "5s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time within which a shard must respond to a write.🎁 What happens if this limit is exceeded? The write goes to hinted handoff?
+ 
+ **write-timeout = "5s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time within which a write operation must complete on the cluster.🎁If this fails is the write rejected? Is this related to the consistency setting or does this timeout happen before or after that consistency setting would apply?
```
jwilder: This is the timeout that the coordinating node must receive a successful response for writing to all remote shard owners before it considers it a failure.  e.g. Node 1 receives a write and it must write it to itself, Node 2 & 3.  This timeout means the local write and the two remote writes must all respond with a success or error or we'll return a timeout error.  This is just a timeout so those writes may still complete successfully, but we stop waiting and queue those writes in hinted handoff and return a error (or partial) write depending on the requested consistency level and number of successful responses received.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+🎁Any notes on what may need to be changed for most systems?
+
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable hinted handoff. 
+
+**dir = "/var/opt/influxdb/hh"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+
+**max-size = 1073741824**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The maximum size of the hinted handoff queue.🎁What happens if this is exceeded? Are older writes pushed off the stack by newer ones? Do new writes for that node fail? Return failures to the client?
```
jwilder: This is the maximum size of a given queue for a node.  If the queue is full, writes to it are rejected and and error is returned to the client.  The queue is only drained when either the writes are retried successfully or they expire.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
 
-_Note: Clustering is in alpha state right now and all clusters __must__ contain three and only three nodes._
+**max-age = "168h"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁Do points age-out after max-age? Do entire files age out? Are they silently dropped? Is anything logged or returned to STDERR?
```
jwilder: `max-age` is how long writes can sit in the queue.  We don't look at the point timestamps, just how long the batch has been in the queue.  When items a purged, we do them at a segment file at a time.  So `max-age` really refers to the length of time a given segment will be allowed to sit on disk.  I don't believe anything is logged when when they expire currently.  Probably a good idea to add that though.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
 
-Configuring a cluster with host machines A, B, and C:
+**retry-rate-limit = 0**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
```
jwilder: `retry-rate-limit`is the rate that hinted handoffs will be retried.  The value indicates rate in bytes per second and applies across all nodes when retried.  For example, `4096` would cause the hinted handoff queue to throttle writes to maintain a throughput of `4kb/s` to each node it has queued writes.   A value of 0 disables the rate limit.  Zero is also the default value.

otoolep: This will change shortly (within a week). The value will be the rate-limit *per node* receiving the traffic.

I would document it as the upcoming meaning.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
 
-1. Install InfluxDB on all three machines.
-2. Generate a config file on each machine by running `influxd config > influxdb.conf`.
-3. Update the `[meta]` section of the configuration file on all three hosts, replacing `localhost` with the hosts actual network IP.
-4. Update the bind-address to another port if 8088 is not acceptable.
-5. On all three machines add `peers = ["A_IP:A_bindaddress", "B_IP:B_bindaddress", "C_IP:C_bindaddress"]` to the `[meta]` section of the config file.
-6. On all three machines, add `replication = 3` in the `[retention]` section of the config file.
-7. Launch `influxd` on hosts A, B, and C in order.
+**retry-interval = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
```
jwilder: This is the initial retry interval that a failed hinted handoff queue write would be retried.  After each failure, it backs off exponentially until it hits the `retry-max-interval`.  It will then retry every `retry-max-interval` until it succeeds.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+
+**election-timeout = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The time that the follower waits for communication from the leader before the follower begins a new election. In practice, InfluxDB staggers this parameter for each 
+Raft peer. The default should work for most systems.
+
+**heartbeat-timeout = "1s"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;The rate at which the leader sends heartbeats to the other Raft peers to preserve its leader status. You may want to alter this parameter depending on your network.
+
+**leader-lease-timeout = "500ms"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
```
benbjohnson: This is the amount of time a raft leader can go without hearing from a majority of nodes before stepping down to a "follower" state.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+
+**commit-timeout = "50ms"**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
```
benbjohnson: This is the amount of time raft will go without issuing a command before a heartbeat is sent to tell the leader it is live.

content/docs/v0.9/administration/config.md
```diff
@@ -2,48 +2,374 @@
...
+**log-enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable logging for continuous query events.
+
+**enabled = true**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;Set this option to `false` to disable continuous queries.
+
+**recompute-previous-n = 2**  
+&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;🎁
+
+**recompute-no-older-than = "10m"**
```
beckettsean: @dgnorton does this setting work in conjunction with the `recompute-previous-n` setting?

For a CQ that runs every 5 minutes, if `recompute-previous-n=2` but `recompute-no-older-than=10m`, does that mean it will only calculate the previous two buckets (5m + 5m = 10m)?

Or does this setting affect some other recalculation parameter?

dgnorton: @beckettsean in pseudo code, `if now() - intervalStartTime > recomputeNoOlderThan then stop`

`now()` and `intervalStartTime` are both wall clock times.  `now() - intervalStartTime` yields a duration (e.g., `8s`) that can be compared to the `recomputeNoOlderThan` duration (e.g. `10s`).  So, for this example, `if 8s > 10s then stop` the expression is `false` and it would not stop.  It would continue and execute the CQ.

One thing I just noticed in the code is how `intervalStartTime` is calculated for the current interval.  It rounds the current time to the nearest interval boundary.  If the current time is 8:08:00 and the `GROUP BY` interval is `15m`, then the start time of the current interval will be rounded to 8:15:00 (i.e., `7m` in the future).

dgnorton: @beckettsean  just talked this over in #core and we shouldn't be rounding to the nearest interval boundary, we should be truncating.  Also, `recompute-no-older-than` needs to be `>= 1` to ensure that every interval gets fully calculated.  I'm going to open an issue for these.

beckettsean:
comments added, looking very very close to ready!

PR #343, opened 2015-09-29
update and reorganize data_exploration

rkuchan:
Made significant changes to the Data Exploration page. I updated the content - all examples refer to the same sample data - and I reorganized the structure.

### DIFF
--- a/content/docs/v0.9/query_language/data_exploration.md
+++ b/content/docs/v0.9/query_language/data_exploration.md
@@ -35,8 +35,8 @@ General tips on query syntax:
 &nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Relative time](../query_language/data_exploration.html#relative-time)  
 &nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Absolute time](../query_language/data_exploration.html#absolute-time)
 * [Regular expressions in queries](../query_language/data_exploration.html#regular-expressions-in-queries)  
-&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and selecting measurements](http://127.0.0.1:1313/docs/v0.9/query_language/data_exploration.html#regular-expressions-and-selecting-measurements)  
-&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and specifying tags](http://127.0.0.1:1313/docs/v0.9/query_language/data_exploration.html#regular-expressions-and-specifying-tags)
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and selecting measurements](..query_language/data_exploration.html#regular-expressions-and-selecting-measurements)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and specifying tags](../query_language/data_exploration.html#regular-expressions-and-specifying-tags)
 
 The examples below query data using [InfluxDB's Command Line Interface (CLI)](../introduction/getting_started.html). For how to directly query data with the HTTP API, see the [Querying Data](../guides/querying_data.html) guide.
 
--- a/content/docs/v0.9/query_language/data_exploration.md
+++ b/content/docs/v0.9/query_language/data_exploration.md
@@ -629,7 +629,7 @@ Return all points that occur after  `2014-01-01 00:00:00`:
 
 Regular expressions are surrounded by `/` characters and use [Golang's regular expression syntax](http://golang.org/pkg/regexp/syntax/). Use regular expressions when selecting measurements and tags.
 
->**Note:** You cannot use regular expressions to specify multiple databases or retention policies.
+>**Note:** You cannot use regular expressions to specify multiple databases, retention policies, or fields.
 
 The [sample data](../query_language/data_exploration.html#sample-data) need to be more intricate for the following sections. Assume that the database `water` now holds several measurements: `h2o_feet`, `h2o_quality`, `h2o_pH`, `average_temperature`, and `h2o_temperature`. Please note that every measurement besides `h2o_feet` contains fictional data.
 
--- a/content/docs/v0.9/query_language/data_exploration.md
+++ b/content/docs/v0.9/query_language/data_exploration.md
@@ -752,7 +752,7 @@ or
 <br>
 <br>
 ```sql
-> SELECT * FROM h2o_feet WHERE location =~ /[y,m]/ AND water_level > 0 LIMIT 4
+> SELECT * FROM h2o_feet WHERE location =~ /[ym]/ AND water_level > 0 LIMIT 4
 ```
 
 CLI response:
--- a/content/docs/v0.9/query_language/data_exploration.md
+++ b/content/docs/v0.9/query_language/data_exploration.md
@@ -5,494 +5,787 @@ aliases:
 ---
 
 
-InfluxDB features an SQL-like query language for querying data and performing aggregations on that data. This section describes the syntax of the query. All queries that return data use the keyword `SELECT`.
+InfluxQL is an SQL-like query language for interacting with data in InfluxDB. The following sections cover useful query syntax for exploring your data.
 
-The primary mechanism for issuing any of the queries listed below is through the HTTP API. For example, the command `SELECT * FROM cpu` can be executed using `curl` as follows:
+The basics:
 
+* [Select a database to query](../query_language/data_exploration.html#select-a-database-to-query)
+* [The `SELECT` statement and the `WHERE` clause](../query_language/data_exploration.html#the-select-statement-and-the-where-clause)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[The basic `SELECT` statement](../query_language/data_exploration.html#the-basic-select-statement)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[The `SELECT` statement and arithmetic](../query_language/data_exploration.html#the-select-statement-and-arithmetic)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[The `WHERE` clause](../query_language/data_exploration.html#the-where-clause)
+* [The `GROUP BY` clause](../query_language/data_exploration.html#the-group-by-clause)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[The basic `GROUP BY` clause](../query_language/data_exploration.html#the-basic-group-by-clause)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[The `GROUP BY` clause and `AS`](../query_language/data_exploration.html#the-group-by-clause-and-as)    
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[The `GROUP BY` clause and `fill()`](../query_language/data_exploration.html#the-group-by-clause-and-fill)  
+
+Limit and sort your results:
+
+* [Limit query returns with `LIMIT` and `SLIMIT`](../query_language/data_exploration.html#limit-query-returns-with-limit-and-slimit)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Limit results per series with `LIMIT`](../query_language/data_exploration.html#limit-the-number-of-results-returned-per-series-with-limit)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Limit the number of series returned with `SLIMIT`](../query_language/data_exploration.html#limit-the-number-of-series-returned-with-slimit)  
+* [Sort query returns with `ORDER BY time DESC`](../query_language/data_exploration.html#sort-query-returns-with-order-by-time-desc)
+* [Paginate query returns with `OFFSET`](../query_language/data_exploration.html#paginate-query-returns-with-offset)
+
+General tips on query syntax:
+
+* [Multiple statements in queries](../query_language/data_exploration.html#multiple-statements-in-queries)
+* [Merge series in queries](../query_language/data_exploration.html#merge-series-in-queries)
+* [Time syntax in queries](../query_language/data_exploration.html#time-syntax-in-queries)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Relative time](../query_language/data_exploration.html#relative-time)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Absolute time](../query_language/data_exploration.html#absolute-time)
+* [Regular expressions in queries](../query_language/data_exploration.html#regular-expressions-in-queries)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and selecting measurements](http://127.0.0.1:1313/docs/v0.9/query_language/data_exploration.html#regular-expressions-and-selecting-measurements)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and specifying tags](http://127.0.0.1:1313/docs/v0.9/query_language/data_exploration.html#regular-expressions-and-specifying-tags)
+
+The examples below query data using [InfluxDB's Command Line Interface (CLI)](../introduction/getting_started.html). For how to directly query data with the HTTP API, see the [Querying Data](../guides/querying_data.html) guide.
+
+#### Sample data
+<br>
+This document uses publicly available data from the [National Oceanic and Atmospheric Administration's (NOAA) Center for Operational Oceanographic Products and Services](http://tidesandcurrents.noaa.gov/stations.html?type=Water+Levels). The data include water levels (ft) collected every six seconds at two stations (Santa Monica, CA (ID 9410840) and Coyote Creek, CA (ID 9414575)) over the period from August 18, 2015 through September 18, 2015.
+
+A subsample of the data in InfluxDB: 
+```
+name: h2o_feet
+----------
+time			               location	     status!				                 water_level
+2015-08-18T00:00:00Z	 santa_monica	 below three feet		          2.064
+2015-08-18T00:00:00Z	 coyote_creek	 between six and nine feet	  8.12
+2015-08-18T00:06:00Z	 coyote_creek	 between six and nine feet	  8.005
+2015-08-18T00:06:00Z	 santa_monica	 below three feet		          2.116
+2015-08-18T00:12:00Z	 coyote_creek	 between six and nine feet	  7.887
+2015-08-18T00:12:00Z	 santa_monica	 below three feet		          2.028
+2015-08-18T00:18:00Z	 coyote_creek	 between six and nine feet	  7.762
+2015-08-18T00:18:00Z	 santa_monica	 below three feet		          2.126
+2015-08-18T00:24:00Z	 coyote_creek	 between six and nine feet	  7.635
+2015-08-18T00:24:00Z	 santa_monica	 below three feet		          2.041
+```
+
+The [series](../concepts/glossary.html#series) are made up of the [measurement](../concepts/glossary.html#measurement) `h2o_feet` and the [tag key](../concepts/glossary.html#tag-key) `location` with the [tag values](../concepts/glossary.html#tag-value) `santa_monica` and `coyote_creek`. There are two [fields](../concepts/glossary.html#field): `water_level` which stores floats and `status!` which stores strings. All of the data are in the database `water`.
+
+> **Disclaimer:** `status!` isn't part of the original NOAA data - we snuck it in there for the sake of having a field key with a special character and string [field values](../concepts/glossary.html#field-value) 😏.
+
+## Select a database to query
+Select the database that you want to work with by entering:
+```sql
+USE <database_name>
+```
+CLI example:
 ```sh
-curl -G 'http://localhost:8086/query' --data-urlencode 'db=mydb' --data-urlencode 'q=SELECT * FROM cpu'
+> USE water
+Using database water
+```
+## The SELECT statement and the `WHERE` clause
+InfluxQL's `SELECT` statement follows the form of an SQL `SELECT` statement where `WHERE` is optional:
+```sql
+SELECT <stuff> FROM <measurement_name> WHERE <some_conditions>
+```  
+
+### The basic `SELECT` statement
+---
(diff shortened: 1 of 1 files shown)


rkuchan:
@jackzampolin, @beckettsean, @gunnaraasen, @mjdesa, @rossmcdonald, and @wfro would love some feedback if you have the time!

jackzampolin:
Ran through this with you and just gave it a re-read.  Looks great.  Love the index up top especially for the long page.

rkuchan:
I should also note that I removed the sections on `Dropping measurements, series, and databases` and `The SHOW command` as I thought those sections belonged on the Schema Exploration page. Happy to add them back in if anyone disagrees.

beckettsean:
@rkuchan totally agree about data vs. schema for the `DROP` and `SHOW` statements.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and specifying tags](../query_language/data_exploration.html#regular-expressions-and-specifying-tags)
 
-Unquoted identifiers must start with an upper or lowercase ASCII letter and can only contain ASCII letters, decimal digits, and the "_" or "." characters.
+The examples below query data using [InfluxDB's Command Line Interface (CLI)](../introduction/getting_started.html). For how to directly query data with the HTTP API, see the [Querying Data](../guides/querying_data.html) guide.
 
-For Example:
+#### Sample data
+<br>
+This document uses publicly available data from the [National Oceanic and Atmospheric Administration's (NOAA) Center for Operational Oceanographic Products and Services](http://tidesandcurrents.noaa.gov/stations.html?type=Water+Levels). The data include water levels (ft) collected every six seconds at two stations (Santa Monica, CA (ID 9410840) and Coyote Creek, CA (ID 9414575)) over the period from August 18, 2015 through September 18, 2015.
```
beckettsean: Can we post the data in line protocol format on S3 so users can download it? We could then provide the `curl` write statement and they could have this data right there with them. Seems pretty easy to me, let's chat.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+
+* Leave identifiers unquoted unless they start with a digit, contain characters other than `[A-z,0-9,_]`, or if they are an [InfluxQL keyword](https://github.com/influxdb/influxdb/blob/master/influxql/INFLUXQL.md#keywords) - then you need to double quote them. Identifiers are database names, retention policy names, user names, measurement names, tag keys, and field keys.
+
+ Select everything from `h2o_feet` by fully qualifying the measurement:
+ ```sql
+ > SELECT * FROM water."default".h2o_feet
+ ``` 
+* Fully qualify a measurement if you wish to query data from a different database or from a retention policy other than the default [retention policy](../concepts/glossary.html#retention-policy). A fully qualified measurement takes the following form:  
+```
+<database>."<retention policy>".<measurement>
```
beckettsean: Should double-quote each of them or none of them. All three are identifiers and potentially must be double-quoted when used like this.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+time
+2015-08-18T00:00:00Z	8.128
+2015-08-18T00:00:00Z	20.24
+[...]
+2015-09-18T21:36:00Z	14.132
+2015-09-18T21:42:00Z	13.876
 ```
 
-## Select and Time Ranges
+> **Note:** When performing arithmetic on fields that store integers be aware that InfluxDB casts those integers to floats.
```
beckettsean: We might want to be explicit that this can lead to overflow errors if the values are above ~2^51. 

"When performing arithmetic on fields that store integers be aware that InfluxDB casts those integers to floats for all mathematical operations. This can lead to [overflow issues](https://influxdb.com/docs/v0.9/troubleshooting/frequently_encountered_issues.html#working-with-really-big-or-really-small-integers) for very large numbers. "

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
-
-Basic arithmetic operations can be performed when querying fields that store floats.
-
+Return data where the tag key `location` has the tag value `santa_monica` and the field `status!` equals `'below three feet'`:
 ```sql
-SELECT value / 2 from cpu_user
+> SELECT * FROM h2o_feet WHERE location = 'santa_monica' AND "status!" = 'below three feet'
+```
+* Always single quote field values that are strings. Note that double quotes do not work when specifying string field values and can cause queries to silently fail.
+* Fields are not indexed so queries on field keys or field values are not performant.
```
beckettsean: I would maybe make this a Note

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+```
+* Always single quote field values that are strings. Note that double quotes do not work when specifying string field values and can cause queries to silently fail.
+* Fields are not indexed so queries on field keys or field values are not performant. 
 
-SELECT (value * 2) + 4 from cpu_user
+More on the `WHERE` clause in InfluxQL:
 
-SELECT (value + 6) / 10 from cpu_user
-```
+* The `WHERE` clause supports comparisons against regular expressions, strings, booleans, floats, integers, and time.
```
beckettsean: We need to be explicit that only the timestamp can be compared to `time`. It is not possible to do something like `WHERE epoch_as_tag > now() - 1d` for example.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
-SELECT (value * 2) + 4 from cpu_user
+More on the `WHERE` clause in InfluxQL:
 
-SELECT (value + 6) / 10 from cpu_user
-```
+* The `WHERE` clause supports comparisons against regular expressions, strings, booleans, floats, integers, and time. 
+* Chain logic together using `AND`  and `OR`, and separate using `(` and `)`.
+* Acceptable comparators include:  
+`=` equal to  
+`<>` not equal to
```
beckettsean: `!=` also works. I wonder if they are equivalant?

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+`<>` not equal to  
+`>` greater than  
+`<` less than  
+`=~` matches against  
+`!~` doesn't match against  
+
+## The GROUP BY clause
 
-## Selecting Multiple Series
+Use the `GROUP BY` clause to group data by tags and/or time intervals. To successfully implement `GROUP BY`,  append the`GROUP BY` clause to a `SELECT` statement and pair the `SELECT` statement with one of InfluxQL's [functions](../query_language/functions.html).
```
beckettsean: `GPOUP BY` **must** come after the `WHERE` clause. We should be explicit, because many SQL implementations don't impose ordering on the subclauses.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+2015-09-18T00:00:00Z	 165
+```
+
+Notice that each timestamp represents a two day interval and that the value in the `count` field is the number of `water_level` points that occurred in that two day interval. You could get the same results by querying the data 17 times - that is, one `COUNT()` query for every two days between August 18, 2015 at midnight and September 18 at 5:00pm - but that could take a while. 
+
+
+Other things to note about `GROUP BY time()`:
+
+* InfluxQL requires a `WHERE` clause if you're using `GROUP BY` with `time()`. Note that unless you specify a different upper and lower bound for the time range, `GROUP BY` uses `epoch 0` as the lower bound and `now()` as the upper bound for the query - this can lead to [unexpected results](../troubleshooting/frequently_encountered_issues.html#getting-unexpected-results-with-group-by-time).
+* Valid time intervals for `time()` are:
```
beckettsean: Maybe "Valid units for `time()` are:"?

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+    `w` weeks   
+
+**GROUP BY tag values and a time interval**  
+Calculate the average `water_level` for the different tag values of `location` in the last two weeks at 6 hour intervals:
+```sql
+> SELECT MEAN(water_level) FROM h2o_feet WHERE time > now() - 2w GROUP BY location,time(6h)
+```
+* Separate multiple `GROUP BY` arguments with a comma.
+
+### The `GROUP BY` clause and `AS`
```
beckettsean: This doesn't feel like anything to do with `GROUP BY`. It's really about the function, not the GROUP BY clause.

Maybe just make this an aside on an earlier query. I don't think it's important to explain it here, could leave it out entirely IMHO

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
 
-> **NOTE:** Regular expressions cannot be used to specify multiple databases or retention policies. Only measurements.
+> **Note:** If you're `GROUP(ing) BY` several things (for example, both tags and a time interval) `fill()` must go at the end of the `GROUP BY` clause.
 
-## Dropping measurements, series, and databases
+## Limit query returns with LIMIT and SLIMIT
+InfluxQL supports two different clauses to limit your query results. Currently, they are mutually exclusive so you may use one or the other, but not both in the same query.
 
-You can drop individual series within a measurement that match given tags, or you can drop entire measurements. Some examples:
+### Limit the number of results returned per series with `LIMIT`
```
beckettsean: We should also document the behavior of `LIMIT` and `SLIMIT` with `GROUP BY *`.

I think we should also mention that the return is not idempotent. Subsequent `LIMIT` queries may return different series depending on the data.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
 ```sql
-SHOW MEASUREMENTS WHERE service = 'redis'
+> SELECT water_level FROM h2o_feet WHERE location = 'santa_monica' ORDER BY time DESC LIMIT 5
 ```
 
-Show series (unique tag sets) on the cpu measurement:
-
-```sql
-SHOW SERIES FROM cpu
+CLI return:
```
beckettsean: hard to see the `08` changing to `09` in the data. Let's be more clear that these are the most recent 5 points.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
-GROUP BY time(10m);
+> SELECT "status!" FROM h2o_feet WHERE time < now() + 1000d
 ```
 
-The time function takes the time interval which can be in microseconds, seconds, minutes, hours, days or weeks. To specify the units you can use the respective suffix `u`, `s`, `m`, `h`, `d` and `w`.
-
-If you issue a query that has an aggregate function like `count` but don't specify a `GROUP BY time` You will only get a single data point back with the number of count from time zero (00:00:00 UTC, Thursday, 1 January 1970).
-
-If you have a `GROUP BY time` clause you should **always** have a `WHERE` clause that limits the scope of time you are looking at. Unless otherwise specified, `GROUP BY` will use `epoch 0` as the lower bound for the time range which is almost never what is desired.
+The other options for specifying time durations with `now()` are listed below. If you do not give a suffix, InfluxDB defaults to microseconds.
```
beckettsean: I think this is incorrect. My testing suggests that in the absence of a unit the entire clause is ignored:

```
> select * from z where time > now() - 1w
> select * from z where time > now() - 1d
> select * from z where time > now() - 1
name: z
-------
time				a	b	c	value
2015-09-16T21:03:08.766228935Z	1	1		42
2015-09-16T21:03:12.05342261Z	1	2		43
2015-09-16T21:03:15.301739075Z	2	2		44
2015-09-16T21:03:18.326026851Z	2	1		45
2015-09-16T21:03:23.214588426Z	2	3		46
2015-09-16T21:03:27.966841551Z	1	3		47
2015-09-16T21:03:32.278902768Z	1	5		48
2015-09-17T23:26:22.051336455Z	2			50
2015-09-17T23:26:23.136249467Z	2			51
2015-09-17T23:26:28.272520389Z		2		55
2015-09-17T23:26:29.608684691Z		2		56
2015-09-17T23:28:30.228423334Z			2	60
2015-09-17T23:28:32.003466798Z			1	60
2015-09-17T23:28:35.315985019Z	2		1	60
2015-09-17T23:28:36.556133935Z	2		1	61
2015-09-17T23:28:37.468202333Z	2		1	63
2015-09-17T23:28:38.316202818Z	2		1	65

> 
```

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+1970-01-01T00:00:00Z	5.359342451341402
 ```
 
-Show 95th percentile of response times in the last day in 30 second intervals:
+> **NOTE:** In InfluxDB, [epoch 0](https://en.wikipedia.org/wiki/Unix_time) (`1970-01-01T00:00:00Z`) is often used as a null timestamp equivalent. If you request a query that has no timestamp to return, such as an aggregation function with an unbounded time range, InfluxDB returns epoch 0 as the timestamp.
+
+## Time syntax in queries  
+InfluxDB is a time series database so, unsurprisingly, InfluxQL has a lot to do with specifying time ranges. If you do not specify start and end times in your query, they default to epoch 0 (`1970-01-01T00:00:00Z`) and `now()`. The following sections detail how to specify different start and end times in queries.
+ 
+### Relative time
```
beckettsean: should we link to the syntax gotcha? It's really easy to do `time > now() -1d` and get a weird error or no result.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+`d` days  
+`w` weeks   
 
-### Filling intervals with no data
-
-By default, `GROUP BY` intervals that have no data will use `null` as the value, by default, though any numerical value, including negative values, are valid values for `fill`. For example, each of the following queries is valid:
+### Absolute time
+---
+**Date time strings**  
+Specify time with date time strings which have the format `YYYY-MM-DD HH:MM:SS.mmm`, where `mmm` are optional milliseconds.
```
beckettsean: Can also give `YYYY-MM-DDTHH:MM:SS.nnnnnnnnnZ`, which is actual RFC3339. Let's check with the engineers on the preferred syntax.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
-```
-```sql
-SELECT COUNT(type) FROM events
-WHERE time > 12345678s
-GROUP BY time(1h) fill(-1)
+> SELECT water_level FROM h2o_feet WHERE time > '2015-08-18 23:00:01.232' AND time < '2015-09-19'
 ```
 
-There are also special options for `fill`. Those values are `null`, `previous`, and `none`. `null` means null is used as the value for intervals without data. `previous` means the values of the previous window is used, and `none` means that all null values are removed. Examples of each are shown below.
+* Single quote the date time string.
```
beckettsean: Worth a note that a double-quoted date won't throw an error, it will just fail to restrict the results at all. (Gets interpreted as a non-existent identifier )

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+* Single quote the date time string. 
+* If you only specify the date, InfluxDB sets the time to `00:00:00`.
 
-```sql
-SELECT COUNT(type) FROM events
-WHERE time > now() - 3h
-GROUP BY time(1h) fill(null)
-```
+**Epoch time**  
+Specify time with timestamps in epoch time. Epoch time is the number of microseconds that have elapsed since 00:00:00 Coordinated Universal Time (UTC), Thursday, 1 January 1970. Indicate the units of the timestamp at the end of the timestamp (see the section above for a list of acceptable time units).
```
beckettsean: Epoch time is the number of **nano**seconds

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
-SELECT COUNT(type) FROM events
-WHERE time > now() - 3h
-GROUP BY time(1h) fill(none)
-```
+## Regular expressions in queries
+
+Regular expressions are surrounded by `/` characters and use [Golang's regular expression syntax](http://golang.org/pkg/regexp/syntax/). Use regular expressions when selecting measurements and tags.
 
-> **Note:** `fill` must go at the end of the group by clause if there are other arguments:
+>**Note:** You cannot use regular expressions to specify multiple databases, retention policies, or fields.
```
beckettsean: "You cannot use regular expressions to match databases, retention policies, or fields. Only measurements and tags can be used with regular expressions."

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
-SELECT value FROM cpu LIMIT 10
+> SELECT * FROM average_temperature,h2o_feet,h2o_pH,h2o_quality,h2o_temperature LIMIT 1
 ```
 
-The following query will return the 10 newest points from each series in the `cpu` measurement, meaning there will be 10 points returned for each unique tag set in `cpu`:
-
+Select the first three points from every measurement whose name starts with `h2o`:  
 ```sql
-SELECT value FROM cpu ORDER BY time DESC LIMIT 10
+> SELECT * FROM /(?i)^h2o\.*/ LIMIT 3
```
beckettsean: I assume `(?i)1 is a case insensitive thing? Seems unnecessary. Also no need to escape the period. I think this is clearer:

`> SELECT * FROM /^h2o.*/ LIMIT 3`

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+---------------------
+time                 degrees	index	location	    pH	water_level
+2015-08-18T00:00:00Z	62            santa_monica
+2015-08-18T00:00:00Z	60            coyote_creek
+2015-08-18T00:06:00Z	60            coyote_creek
 ```
 
-The following query will return the 10 newest points less than an hour old in the `cpu` measurement, regardless of the tag set, meaning there will be at most 10 points returned:
-
+* Alternatively, you could have gotten this return by specifying the retention policy `default`:
```
beckettsean: What is this supposed to demonstrate? Was there originally more about multiple RPs in this doc?

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+2015-08-18T00:00:00Z	coyote_creek	8.12
+2015-08-18T00:06:00Z	coyote_creek	8.005
+2015-08-18T00:12:00Z	coyote_creek	7.887
+2015-08-18T00:18:00Z	coyote_creek	7.762
+```
 
+Select the oldest four points from the measurement `h2o_feet` where the value of the tag `location` includes a `y` or an `m` and `water_level` is greater than zero:
 ```sql
-SELECT * FROM cpu WHERE host='server-a' LIMIT 10 OFFSET 10
+> SELECT * FROM h2o_feet WHERE (location =~ /.*y.*/ OR location =~ /.*m.*/) AND water_level > 0 LIMIT 4
```
beckettsean: I think this can also be written as:

`> SELECT * FROM h2o_feet WHERE location =~ /[y,m]/ AND water_level > 0 LIMIT 4`

If so, it's worth giving both examples

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Relative time](../query_language/data_exploration.html#relative-time)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Absolute time](../query_language/data_exploration.html#absolute-time)
+* [Regular expressions in queries](../query_language/data_exploration.html#regular-expressions-in-queries)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and selecting measurements](..query_language/data_exploration.html#regular-expressions-and-selecting-measurements)  
+&nbsp;&nbsp;&nbsp;◦&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;[Regular expressions and specifying tags](../query_language/data_exploration.html#regular-expressions-and-specifying-tags)
 
-Unquoted identifiers must start with an upper or lowercase ASCII letter and can only contain ASCII letters, decimal digits, and the "_" or "." characters.
+The examples below query data using [InfluxDB's Command Line Interface (CLI)](../introduction/getting_started.html). For how to directly query data with the HTTP API, see the [Querying Data](../guides/querying_data.html) guide.
```
gunnaraasen: Second sentence sounds awkward. Maybe reword it as:
> Data can also be queried directly using the HTTP API as shown in the [Querying Data](../guides/querying_data.html) guide.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+2015-08-18T00:24:00Z	 coyote_creek	 between six and nine feet	  7.635
+2015-08-18T00:24:00Z	 santa_monica	 below three feet		          2.041
 ```
 
-To select for a `cpu` measurement with a tag key of `cpu-1` and tag value of `1`:
-```sql
-SELECT * FROM cpu
-WHERE "cpu-1" = '1'
-```
+The [series](../concepts/glossary.html#series) are made up of the [measurement](../concepts/glossary.html#measurement) `h2o_feet` and the [tag key](../concepts/glossary.html#tag-key) `location` with the [tag values](../concepts/glossary.html#tag-value) `santa_monica` and `coyote_creek`. There are two [fields](../concepts/glossary.html#field): `water_level` which stores floats and `status!` which stores strings. All of the data are in the database `water`.
```
gunnaraasen: Two small nits:
- First, "All of the data are in the database `water`." -> "All of the data is stored in the `water` database."
- Second, can the `water` database be renamed to something more descriptive like `NOAA_water_levels`?

content/docs/v0.9/query_language/data_exploration.md
gunnaraasen: I just fired off an email to the NOAA requesting a copy of the 1-minute resolution data.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
 
-## Selecting the Database and Retention Period
-When selecting data using the query language, the target database and retention period can optionally be specified. Doing so is known as "fully qualifying" your measurement. A fully-qualified measurement is in the following form:
+> **Disclaimer:** `status!` isn't part of the original NOAA data - we snuck it in there for the sake of having a field key with a special character and string [field values](../concepts/glossary.html#field-value) 😏.
```
gunnaraasen: The `status!` name feels a little forced with the exclamation point at the end. Maybe something like `water level description` would be better (with spaces as the special characters).

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
 
+## Select a database to query
+Select the database that you want to work with by entering:
```
gunnaraasen: Not sure if this needs to be it's own section since the `USE` syntax is specific to the CLI. I would tack this on at the end of the "sample data" section with a quick
> Don't forget to set the database in the CLI using `USE water` before running the queries in this guide. All data exploration queries require a database to be set.

content/docs/v0.9/query_language/data_exploration.md
```diff
@@ -5,494 +5,787 @@ aliases:
...
+USE <database_name>
 ```
-"<database>"."<retention period>".<measurement>
+CLI example:
+```sh
+> USE water
+Using database water
 ```
+## The SELECT statement and the `WHERE` clause
+InfluxQL's `SELECT` statement follows the form of an SQL `SELECT` statement where `WHERE` is optional:
```
gunnaraasen: "where `WHERE` is optional" -> "where the `WHERE` clause is optional"

rkuchan:
Data download to come but all other edits are done.

PR #346, opened 2015-09-30
add show databases and restructure schema_exploration.md

rkuchan:
I restructured schema_exploration.md and all the examples now use the same database as the newer data_exploration.md.

There's also a new section on `SHOW DATABASES`.

### DIFF
--- a/content/docs/v0.9/query_language/schema_exploration.md
+++ b/content/docs/v0.9/query_language/schema_exploration.md
@@ -14,7 +14,7 @@ The examples below query data using [InfluxDB's Command Line Interface (CLI)](..
 
 **Sample data**  
 
-This document uses the same sample data as the [Data Exploration](../query_language/data_exploration.html) page. Note that some of the measurements in the data are fictional - they're meant to make the next sections more explanatory and (hopefully) more interesting. The next sections will get you acquainted with the schema of the sample data in the database `water`.
+This document uses the same sample data as the [Data Exploration](../query_language/data_exploration.html) page. Note that some of the measurements and data in the database are fictional - they're meant to make the next sections more explanatory and (hopefully) more interesting. The next sections will get you acquainted with the schema of the sample data in the `NOAA_water_database` database.
 
 ## See all databases with `SHOW DATABASES`
 Get a list of all the databases in your system by entering:
@@ -28,7 +28,7 @@ CLI example:
 name: databases
 ---------------
 name
-water
+NOAA_water_database
 ```
 
 ## Explore series with `SHOW SERIES`
@@ -38,12 +38,12 @@ The `SHOW SERIES` query returns the distinct [series](../concepts/glossary.html#
 SHOW SERIES FROM <measurement_name> WHERE <tag_key>=<'tag_value'>
 ```
 
-Return all series in the database `water`:
+Return all series in the database `NOAA_water_database`:
 ```sql
 > SHOW SERIES
 ```
 
-CLI return:  
+CLI response:  
 ```sh
 name: average_temperature
 -------------------------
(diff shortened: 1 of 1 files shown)


content/docs/v0.9/query_language/schema_exploration.md
```diff
@@ -2,251 +2,254 @@
...
+SHOW SERIES FROM <measurement_name> WHERE <tag_key>=<'tag_value'>
+```
 
+Return all series in the database `water`:
 ```sql
-SHOW MEASUREMENTS
+> SHOW SERIES
+```
+
+CLI return:
```
otoolep: CLI "response"?

content/docs/v0.9/query_language/schema_exploration.md
```diff
@@ -2,251 +2,254 @@
...
-SHOW MEASUREMENTS
+> SHOW SERIES
+```
+
+CLI return:  
+```sh
+name: average_temperature
+-------------------------
+_key						                                   location
+average_temperature,location=coyote_creek	   coyote_creek
```
otoolep: Can we line all this up?

otoolep:
Good content, some minor feedback.

rkuchan:
Thanks! I changed all the `CLI return`s to `CLI response`. It doesn't look like it here but the data in the CLI responses are all lined up for the website's format.

beckettsean:
It feels weird to have all the SHOW queries here and none of the DROP queries. Is that a considered choice or just because they didn't used to be here?

content/docs/v0.9/query_language/schema_exploration.md
```diff
@@ -2,251 +2,254 @@
...
+```
 
+From the output you can see that the database `NOAA_water_database` has five measurements: `average_temperature`, `h2o_feet`, `h2o_pH`, `h2o_quality`, and `h2o_temperature`.
+
+Return measurements where the tag key `randtag` equals `1`:
 ```sql
-SHOW SERIES
-FROM cpu_load
-WHERE region = 'uswest'
+> SHOW MEASUREMENTS WHERE randtag = '1'
```
beckettsean: Feels like one of the examples with a tag filter should use a regex instead of a string match, just to show it's possible.

beckettsean: Maybe `SHOW MEASUREMENTS WHERE randtag =~ /\d/` to show regex matching?

content/docs/v0.9/query_language/schema_exploration.md
```diff
@@ -2,251 +2,254 @@
...
+
+## Explore tag values with SHOW TAG VALUES
+The `SHOW TAG VALUES` query returns the set of [tag values](../concepts/glossary.html#tag-value) for a specific tag key across all measurements in the database. It takes the following form, where the `FROM` clause is optional:
 ```sql
-SHOW TAG VALUES
-FROM cpu_load
-WITH KEY=host
+SHOW TAG VALUES FROM <measurement_name> WITH KEY = tag_key
 ```
+> **Note:** The tag_key is **not** quoted.
```
beckettsean: It is not single-quoted like strings would be, but as an identifier it might need to be double-quoted. I would remove this note, I think the actual examples make it clear the value should not be single-quoted.

beckettsean:
looks great, just a few minor notes and one big question.

rkuchan:
Got rid of the Note and added a regular expression example to the `SHOW MEASUREMENTS` section. 

I didn't include the `DROP` queries because I thought they might belong in administration? I can add them in though - they were in data_exploration.md initially.

rkuchan:
https://github.com/influxdb/influxdb.com/issues/355

beckettsean:
:+1:

PR #352, opened 2015-10-01
Update Chronograf pages for v0.2.0

toddboom:


toddboom:
Removing lots of grossness, should be relatively decent. Self-approving and merging.

PR #354, opened 2015-10-01
adding clustering instructions for Ubuntu 15.04

jackzampolin:
@beckettsean @rossmcdonald

### DIFF
--- a/content/docs/v0.9/guides/clustering.md
+++ b/content/docs/v0.9/guides/clustering.md
@@ -28,25 +28,13 @@ Throughout this example, each node will be given a number that denotes the order
 1. Install InfluxDB on the 3 machines following the [installation guide](/docs/v0.9/introduction/installation.html). Do not start the daemon on any of the machines.
 2. For each node's `/etc/opt/influxdb/influxdb.conf` file, replace `hostname = "localhost"` with your host's actual name. This hostname must be resolved by all members in the cluster. It can be an IP or a hostname and optional port number if necessary.
 3. For each node's `/etc/opt/influxdb/influxdb.conf` file, update the bind-address to another port if `8088` is unacceptable. The bind-address can also specify the host interface IP to use (e.g. `10.0.1.10:8088`). By default it will bind on all interfaces. Note that the port may differ from node to node (e.g. one can use `8088`, another use `9099`, and the other `10101`).
-4. Start InfluxDB on the first node, `sudo service influxdb start`.
-5. In `/etc/init.d/influxdb` on the second node, set `INFLUXD_OPTS="-join hostname_1:port_1"`.
-6. Start InfluxDB on the second node, `sudo service influxdb start`.
-7. In `/etc/init.d/influxdb` on the third node, set `INFLUXD_OPTS="-join hostname_1:port_1,hostname_2:port_2"`.
-8. Start InfluxDB on the third node, `sudo service influxdb start`.
+4. Start InfluxDB on the first node, `sudo service influxdb start` (or `sudo systemctl start influxdb` if you are using systemd).
+5. In `/etc/init.d/influxdb` on the second node, set `INFLUXD_OPTS="-join hostname_1:port_1"`.  If using systemd this command must be set in the `/etc/default/influxdb` file.  If it does not exist, create it.
+6. Start InfluxDB on the second node, `sudo service influxdb start` (or `sudo systemctl start influxdb` if you are using systemd).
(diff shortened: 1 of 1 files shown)


beckettsean:
@mjdesa please review, also

rossmcdonald:
I'm wondering if it wouldn't make more sense here to avoid re-copying the section over again, and instead do something like this:

> 4- Start InfluxDB on the first node, `sudo service influxdb start` (or `sudo systemctl start influxdb` if you are using systemd).

Where the new commands are just added to the end. That way we avoid copying the same section over again just to add different commands.

rossmcdonald:
Also, it looks like Ubuntu is trying to ease the upgrade process by having `service` be an alias for `systemctl`, so the commands are actually the same on Ubuntu 15.04:

```
vagrant@influx1:~$ service influxdb status
● influxdb.service - InfluxDB is an open-source, distributed, time series database
   Loaded: loaded (/lib/systemd/system/influxdb.service; enabled; vendor preset: enabled)
   Active: active (running) since Fri 2015-10-02 14:32:06 UTC; 2min 29s ago
     Docs: https://influxdb.com/docs/
 Main PID: 2262 (influxd)
   CGroup: /system.slice/influxdb.service
           └─2262 /opt/influxdb/influxd -config /etc/opt/influxdb/influxdb.conf
vagrant@influx1:~$ systemctl status influxdb
● influxdb.service - InfluxDB is an open-source, distributed, time series database
   Loaded: loaded (/lib/systemd/system/influxdb.service; enabled; vendor preset: enabled)
   Active: active (running) since Fri 2015-10-02 14:32:06 UTC; 3min 26s ago
     Docs: https://influxdb.com/docs/
 Main PID: 2262 (influxd)
   CGroup: /system.slice/influxdb.service
           └─2262 /opt/influxdb/influxd -config /etc/opt/influxdb/influxdb.conf
```

I don't think any of the other distributions are doing this, though, so RedHat/CentOS 7 users will have to use the `systemctl` command. Just something to be aware of.

jackzampolin:
@rossmcdonald I was having trouble with markdown and the list numbering.  Looks like when you start an ordered list in markdown it always starts at 1.  I'll look for a work around again today but didn't find one when I checked yesterday.

content/docs/v0.9/guides/clustering.md
```diff
@@ -25,17 +25,29 @@ The following is the current recommended procedure for configuring a cluster.
...
 
-1. Install InfluxDB on the 3 machines.
+1. Install InfluxDB on the 3 machines following the [installation guide](/docs/v0.9/introduction/installation.html). Do not start the daemon on any of the machines.
 2. For each node's `/etc/opt/influxdb/influxdb.conf` file, replace `hostname = "localhost"` with your host's actual name. This hostname must be resolved by all members in the cluster. It can be an IP or a hostname and optional port number if necessary.
 3. For each node's `/etc/opt/influxdb/influxdb.conf` file, update the bind-address to another port if `8088` is unacceptable. The bind-address can also specify the host interface IP to use (e.g. `10.0.1.10:8088`). By default it will bind on all interfaces. Note that the port may differ from node to node (e.g. one can use `8088`, another use `9099`, and the other `10101`).
-4. Start InfluxDB on the first node.
+4. Start InfluxDB on the first node, `sudo service influxdb start`.
```
rossmcdonald: @jackzampolin Sorry, I should have been more clear. My suggestion would be to put both commands on this line instead of duplicating the section below just to add systemd commands. So this line would become:

> Start InfluxDB on the first node, `sudo service influxdb start` (or `sudo systemctl start influxdb` if you are using systemd).

And the second section (lines 39-50) can be dropped altogether, since it's mostly repeating the first section.

Does that make sense?

beckettsean:
@rossmcdonald that's a great suggestion

jackzampolin:
@rossmcdonald That is a great suggestion!  Thanks!

mjdesa:
:+1:

beckettsean:
@jackzampolin the larger doc still mentions 0.9.3 a lot. Let's remove those in favor of something more future-proof.

jackzampolin:
@beckettsean 0.9.3 is mentioned because there were a number of breaking changes to how clustering works introduced with that release.  Would you still like me to remove them?

beckettsean:
@jackzampolin I went ahead and made the changes I thought it needed, please review https://github.com/influxdb/influxdb.com/commit/8ce81ddd5f98dce13dab0bec119d641431da7241 and let me know if that makes sense.

jackzampolin:
@beckettsean Think it looks great! Thanks! :+1:

PR #356, opened 2015-10-02
update chronograf downloads

rkuchan:


gunnaraasen:
Whoops, that was my mistake. Thanks @rkuchan!

Issue #307, opened 2015-09-11
explain that InfluxDB is not CRUD, just CRd


beckettsean:
Create
Read
Update
Delete

InfluxDB only does two of those fully, and has partial support for DELETE. Let's add a section to the SQL comparison page that differentiates InfluxQL from SQL in this way. Basically UPDATE doesn't exist but there is overwrite. DELETE is only for big chunks of data, at least an entire series. Link to the open issue to have DELETE SERIES support a WHERE time clause.

jackzampolin:
https://github.com/influxdb/influxdb.com/pull/358

Issue #306, opened 2015-09-11
Add prominent docs in the API intro disabusing people of the notion of "HTTP" and "REST"


pauldix:
For some reason a number of people really get hung up on the idea that HTTP is simply a transport for the InfluxDB API and that its semantics are not used in any way. We should clearly document this.

The reasoning for using HTTP was simple: there are easy to use libraries in every language to make HTTP requests, making it an ideal candidate for sending InfluxQL requests to the database. However, it is only a transport.

See https://twitter.com/Kozure1024/status/642360310142099457 for example of the confusion.

beckettsean:
Good idea, @pauldix, and reminds me to explain that InfluxQL [isn't CRUD](https://github.com/influxdb/influxdb.com/issues/307), either.

beckettsean:
Put this on the Writing Data guide just before the HTTP Response Codes section.

jackzampolin:
https://github.com/influxdb/influxdb.com/pull/358

PR #358, opened 2015-10-02
CRUD and REST

jackzampolin:
This is in response to the below issues:
https://github.com/influxdb/influxdb.com/issues/306
https://github.com/influxdb/influxdb.com/issues/307


content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -142,3 +142,12 @@ InfluxQL also supports regular expressions, arithmetic in expressions, `SHOW` st
 
 Now that you have the general idea, check out our [Getting Started Guide](https://influxdb.com/docs/v0.9/introduction/getting_started.html).
 
+## A note on why InfluxDB isn't CRUD...
+
+InfluxDB is a time series database and there is not a need for full UPDATE or DESTROY functionality.  It is optimized to store high volume, time series data.  This data commonly comes from sources like distributed sensor groups, click data from large websites, lists of financial transactions etc...
```
beckettsean: I'd be more judicious here. There are many reasonable _needs_, but in order to meet the more common needs of the community we have to make assumptions. Assuming that deletes and updates are rare or impossible allows for the system to be more performant with reads and writes.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -142,3 +142,12 @@ InfluxQL also supports regular expressions, arithmetic in expressions, `SHOW` st
...
+
+What this data has in common is that it is much more useful in the aggregate.  One reading saying that your computer’s CPU is at 12% utilization at 12:38:35 on a Tuesday is useless on it’s own.  It only becomes useful when it is paired with the rest of the series and visualized.  This is where trends over time begin to show, and actionable insight can be drawn from the data.
```
beckettsean: again I'd qualify this a bit. It's not useless, it's just not as useful, generally. There are many cases even in devops where a single outlier point is interesting in and of itself, but the database isn't built for the special cases, it's built for the 80% case.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -142,3 +142,12 @@ InfluxQL also supports regular expressions, arithmetic in expressions, `SHOW` st
...
 
+## A note on why InfluxDB isn't CRUD...
+
+InfluxDB is a time series database and there is not a need for full UPDATE or DESTROY functionality.  It is optimized to store high volume, time series data.  This data commonly comes from sources like distributed sensor groups, click data from large websites, lists of financial transactions etc...
+
+What this data has in common is that it is much more useful in the aggregate.  One reading saying that your computer’s CPU is at 12% utilization at 12:38:35 on a Tuesday is useless on it’s own.  It only becomes useful when it is paired with the rest of the series and visualized.  This is where trends over time begin to show, and actionable insight can be drawn from the data.  
+
+Time series also shares the attribute that it is generally written once and never updated.  These observations drive the design and development of InfluxDB.  Updating in InfluxDB is actually an overwrite operation.  This functionality is slow, but allows for batch updating: e.g. a sensor was shown to be consistently wrong by 2° C.  Destroy similarly works only on batches of data: e.g. a whole series or measurement.
```
beckettsean: > This functionality is slow
I actually don't know that this is true. Overwriting a point is no different from writing it in the first place. The slowness comes more from writing out-of-order points. We should talk with Paul about this.

> but allows for batch updating
Really? I'm not sure I follow.

> a sensor was shown to be consistently wrong by 2° C.  
feels like the start of an example but there's no payoff. For a blog post the tone would be okay but for docs we need to give the complete picture.

corylanou: Updating can be as slow as inserting out of order from what I understand.  Mostly because in the new storage engine it is read only trees, so you have to re-encode the entire block at a future point to fully integrate that new update in the block.  No different than writing an out of order point.

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -142,3 +142,12 @@ InfluxQL also supports regular expressions, arithmetic in expressions, `SHOW` st
...
+
+The result is that InfluxDB is not a full CRUD database but more like a CR-ud, creating and reading data at performant speeds while having limited update and destroy ability.  For more information on why InfluxDB made these architectural decisions [Paul Dix](https://github.com/pauldix) has an excellent [blog post](https://influxdb.com/blog/2015/06/03/InfluxDB_clustering_design.html) with more explanation.
```
beckettsean: > creating and reading data at performant speeds while having limited update and destroy ability.

maybe re-word to something like:

The result is that InfluxDB is not a full CRUD database but more like a CR-ud, prioritizing the performance of creating and reading data over update and destroy, and preventing some update and destroy behaviors to make create and read more performant.

content/docs/v0.9/guides/writing_data.md
```diff
@@ -58,6 +58,15 @@ Write the data in `cpu_data.txt` to the `mydb` database with:
 ---
 InfluxDB is a schemaless database. You can add new measurements, tags, and fields at any time. Note that if you attempt to write data with a different type than previously used (for example, writing a string to a field that previously accepted integers), InfluxDB will reject those data.
 
+### A note on REST...
+---
+InfluxDB uses HTTP solely as a convenient and widely supported data transfer language.  
+
+Why are modern web API’s constructed in a REST style?  Because as the number of endpoints grows the need for an organizing system becomes pressing.  REST is the industry agreed style for organizing large numbers of endpoints.  This consistency is good for programmers who are both developing and consuming the API: everyone involved knows what to expect.
```
beckettsean: strike the comma in "APIs", it's neither a contraction nor a possessive.

I also don't like asking leading questions except in a socratic dialog, which this is not. In addition we don't want to make a lot of statements about what is and isn't true about REST, that's like stating that `vim` is better than `emacs` in our docs. Perhaps instead:

"Modern web APIs have settled on the REST convention because it addresses a common need. As the number of endpoints grows the need for an organizing system becomes pressing.  REST is the industry agreed style for organizing large numbers of endpoints.  This consistency is good for those developing and consuming the API: everyone involved knows what to expect.

content/docs/v0.9/guides/writing_data.md
```diff
@@ -58,6 +58,15 @@ Write the data in `cpu_data.txt` to the `mydb` database with:
 ---
 InfluxDB is a schemaless database. You can add new measurements, tags, and fields at any time. Note that if you attempt to write data with a different type than previously used (for example, writing a string to a field that previously accepted integers), InfluxDB will reject those data.
 
+### A note on REST...
+---
+InfluxDB uses HTTP solely as a convenient and widely supported data transfer language.  
+
+Why are modern web API’s constructed in a REST style?  Because as the number of endpoints grows the need for an organizing system becomes pressing.  REST is the industry agreed style for organizing large numbers of endpoints.  This consistency is good for programmers who are both developing and consuming the API: everyone involved knows what to expect.  
+
+REST, however, is just a convention.  InfluxDB makes due with 3 API endpoints.  This simple, easy to understand system just uses HTTP as a transfer system for [InfluxQL](https://github.com/influxdb/influxdb/blob/master/influxql/INFLUXQL.md).
```
beckettsean: A note that the docs have extra whitespace throughout. Our style guide doesn't use two spaces before a sentence.

jackzampolin:
@beckettsean Cleaned up the language a bit.  See what you think!

content/docs/v0.9/concepts/crosswalk.md
```diff
@@ -142,3 +142,10 @@ InfluxQL also supports regular expressions, arithmetic in expressions, `SHOW` st
 
 Now that you have the general idea, check out our [Getting Started Guide](https://influxdb.com/docs/v0.9/introduction/getting_started.html).
 
+## A note on why InfluxDB isn't CRUD...
+
+InfluxDB is a database that has been optimized for time series data. This data commonly comes from sources like distributed sensor groups, click data from large websites, or lists of financial transactions.
+
+One thing this data has in common is that it is more useful in the aggregate. One reading saying that your computer’s CPU is at 12% utilization at 12:38:35 UTC on a Tuesday is hard to draw conclusions from. It becomes more useful when it is paired with the rest of the series and visualized. This is where trends over time begin to show, and actionable insight can be drawn from the data. Time series also shares the attribute that it is generally written once and rarely updated.
```
beckettsean: nit: change "...paired..." to "...more useful when it is combined with the rest..."

beckettsean: "Time series also shares the attribute that it is generally written once and rarely updated."

to

"In addition, time series data is generally written once and rarely updated."

content/docs/v0.9/guides/writing_data.md
```diff
@@ -58,6 +58,15 @@ Write the data in `cpu_data.txt` to the `mydb` database with:
 ---
 InfluxDB is a schemaless database. You can add new measurements, tags, and fields at any time. Note that if you attempt to write data with a different type than previously used (for example, writing a string to a field that previously accepted integers), InfluxDB will reject those data.
 
+### A note on REST...
+---
+InfluxDB uses HTTP solely as a convenient and widely supported data transfer language.
```
beckettsean: data transfer protocol?

content/docs/v0.9/guides/writing_data.md
```diff
@@ -58,6 +58,15 @@ Write the data in `cpu_data.txt` to the `mydb` database with:
 ---
 InfluxDB is a schemaless database. You can add new measurements, tags, and fields at any time. Note that if you attempt to write data with a different type than previously used (for example, writing a string to a field that previously accepted integers), InfluxDB will reject those data.
 
+### A note on REST...
+---
+InfluxDB uses HTTP solely as a convenient and widely supported data transfer language.  
+
+Modern web APIs have settled on REST because it addresses a common need. As the number of endpoints grows the need for an organizing system becomes pressing. REST is the industry agreed style for organizing large numbers of endpoints. This consistency is good for those developing and consuming the API: everyone involved knows what to expect.
+
+REST, however, is just a convention. InfluxDB makes due with 3 API endpoints. This simple, easy to understand system just uses HTTP as a transfer system for [InfluxQL](https://github.com/influxdb/influxdb/blob/master/influxql/INFLUXQL.md).
```
beckettsean: strike "just". 

spell out "three"

Also let's add one final sentence like: "The InfluxDB API makes no attempt to be RESTful."

beckettsean:
:+1: merge it!

Issue #251, opened 2015-08-25
give service start instructions for OS X


beckettsean:
From https://github.com/influxdb/influxdb/issues/3826#issuecomment-134709778

beckettsean:
Homebrew installtion gives the following directions. We should make sure the two don't contradict each other:

```
$ brew install influxdb
==> Downloading https://homebrew.bintray.com/bottles/influxdb-0.9.3.yosemite.bottle.tar.gz
######################################################################## 100.0%
==> Pouring influxdb-0.9.3.yosemite.bottle.tar.gz
==> Caveats
Config files from old InfluxDB versions are incompatible with version 0.9.
If upgrading from a pre-0.9 version, the new configuration file will be named:
    /usr/local/etc/influxdb.conf.default
To generate a new config file:
    influxd config > influxdb.conf.generated

To have launchd start influxdb at login:
  ln -sfv /usr/local/opt/influxdb/*.plist ~/Library/LaunchAgents
Then to load influxdb now:
  launchctl load ~/Library/LaunchAgents/homebrew.mxcl.influxdb.plist
Or, if you don't want/need launchctl, you can just run:
  influxd -config /usr/local/etc/influxdb.conf
==> Summary
🍺  /usr/local/Cellar/influxdb/0.9.3: 4 files, 28M
```

jackzampolin:
@rossmcdonald Are those launchd instructions current?  Any advice here?

rossmcdonald:
@jackzampolin Those will work for OS X Yosemite. I'm not 100% sure if those are still correct for El Capitan, though.

jackzampolin:
@rossmcdonald Thanks! Going to note that in the docs!

jackzampolin:
closed by https://github.com/influxdb/influxdb.com/pull/362

PR #362, opened 2015-10-06
issue#251

jackzampolin:
https://github.com/influxdb/influxdb.com/issues/251


beckettsean:
There's a lot more to the instructions spit out by the brew recipe: https://github.com/influxdb/influxdb.com/issues/251#issuecomment-135576044

I think these are worth repeating/explaining, so that InfluxDB is being managed by the OS X service manager rather than just a daemon launched from shell.

beckettsean:
:+1:

PR #378, opened 2015-10-08
fix install instructions for ubuntu/debian and CENTOS

jackzampolin:
fixes https://github.com/influxdb/influxdb/issues/2933


beckettsean:
@jackzampolin I think this trades one problem for another. As I understand it, the instructions for a 14.04 package are incompatible with a 15.04 package, and vice versa. We should provide both, since both are in use.

@rossmcdonald your thoughts?

rossmcdonald:
Calling `/etc/default/influxdb start` won't start the InfluxDb service, so I don't think this solves the problem.

It sounds like the main issue with influxdb/influxdb#2923 is that the `influx` binary is not placed in the user path. There are really two ways to fix this (from a documentation perspective):
* Add it to the user path by running `export PATH=$PATH:/opt/influxb`
* Symlink the binary to a location which is already on the user path (such as `/usr/bin`). This can be accomplished by running `ln -s /opt/influxdb/influx /usr/bin/influx`

I think the second option is easiest. I'll also make a PR to have the installation script do this so that just calling `influx` works out of the box for future releases.

beckettsean:
@rossmcdonald @jackzampolin I think we close this then, in favor of Ross's install PR.

jackzampolin:
@sean was thinking this too.

Issue #228, opened 2015-08-17
document comments in write protocol


beckettsean:
`#` at the beginning of the line is a valid comment character for the line protocol. All subsequent characters are ignored.

PR #379, opened 2015-10-08
issue #228

jackzampolin:
fixes https://github.com/influxdb/influxdb.com/issues/228


beckettsean:
:+1: although lots of whitespace changes. I think Atom is a bit too aggressive about that.

PR #393, opened 2015-10-13
rp not valid query string param for reads

### LINKED ISSUE #390: Retention policy not supported as URL query param for queries
https://influxdb.com/docs/v0.9/query_language/query_syntax.html#retention-policy

This is incorrect. `rp` is only supported for writes, not queries.

https://github.com/influxdb/influxdb/issues/4413

beckettsean:
fixes #390


otoolep:
+1

PR #406, opened 2015-10-18
Add Week In Review post For 10/19.

toddboom:


toddboom:
Ready to rock, self-approval granted.

Issue #412, opened 2015-10-20
"authentication page" link in "Querying Data" is broken


malteo:
There's a broken link to https://influxdb.com/docs/v0.9/administration/authentication.html from the https://influxdb.com/docs/v0.9/guides/querying_data.html page

PR #416, opened 2015-10-20
fix links to authentication_and_authorization.md

rkuchan:
Correct aliases in authentication_and_authorization.md - my bad.

Fixes: https://github.com/influxdb/influxdb.com/issues/412

Thanks, @malteo for catching this!


gunnaraasen:
:+1:

Issue #298, opened 2015-09-09
Service start instructions need tweak fro Ubuntu 15


otoolep:
With Ubuntu 15, one needs to do `systemctl start influxdb` to start the service.

See https://github.com/influxdb/influxdb/issues/3989

rossmcdonald:
Updated here: #419

PR #419, opened 2015-10-20
Updated systemd installation page

rossmcdonald:
Few changes to the Installation page, namely:
* I updated the language flag on the markdown code samples to be `shell` instead of `bash` or `sh`. My only goal here was to make the flags consistent.
* I updated the `start` commands to use `service` instead of `/etc/init.d/influxdb`. This ensures that these commands are able to be used on systemd distributions without issue. #298 
* I changed the "Hosting on AWS" section to point to modifying the `/etc/default/influxdb` file instead of `/etc/init.d/influxdb` (for systemd). I also added a section on permissions, since the original section relied on modifying the init script.

The goal with these changes was to make all commands agnostic to the init system.


gunnaraasen:
LFTM

beckettsean:
:+1:

PR #420, opened 2015-10-20
Initial docs for kapacitor 

nathanielc:
There wasn't a clear spot to put all this new documentation so I put it where it made sense to me. We can move it around if there is a better place. I only had to modify two files from the original docs to get it to work. Obviously nothing links to this yet and thats fine.

There is still some more docs that need to be written but thought I'd get a head start on the PR.


gunnaraasen:
The placement of the files is fine for now. New docs will have a real home for kapacitor.

I put this branch up on the staging site. I'll finish looking this over tonight.

gunnaraasen:
https://new.influxdb.com/docs/kapacitor/v0.1/introduction/getting_started.html

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,382 @@
...
+
+Kapacitor is a data processing engine. It can process both stream and batch data.
+This guide will walk you through both workflows and teach you the basics of using 
+and running a Kapacitor daemon.
+
+
+What you will need
+------------------
+
+* InfluxDB - While Kapacitor does not require InfluxDB it is the easiest to setup and so we will use it in this guide.
```
gunnaraasen: Link to the InfluxDB installation guide.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,382 @@
...
+    You will need InfluxDB >= 0.9.5
+* Telegraf  - We will use a specific Telegraf config to send data to InfluxDB so that the examples Kapacitor tasks have context.
```
gunnaraasen: Ditto

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,382 @@
...
+    You will need Telegraf >= 0.1.9 since the names of measurements change prior to that version.
+* Kapacitor - You can get the latest Kapacitor binaries for your OS at the [downloads](/download/#download) page.
+* Terminal - Kapacitor's interface is via a CLI and so you will need a basic terminal to issue commands.
+
+
+The Use Case
+------------
+
+For the guide we will follow the classic use case of triggering an alert for high cpu usage on a server.
+It is simple use case but will make learning Kapacitor easy since we do not have to think about the complexity
```
gunnaraasen: Not sure you need to comment on the simplicity of the example here. Every getting started guide uses a simple example.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+Installing Kapacitor
+--------------------
+
+Kapacitor has two binaries:
+
+* kapacitor -- a CLI program for calling the Kapacitor API.
+* kapacitord -- the Kapacitor server daemon.
+
+You can either download the binaries directly from the [downloads](/download/#download) page or `go get` them:
```
gunnaraasen: This should be moved to an installation page.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+```
+
+Put the above configuration in a file called `telegraf.conf` and start telegraf:
+
+```sh
+telegraf -config telegraf.conf
+```
+
+OK, at this point we should have a running InfluxDB + Telegraf setup.
+There should be some cpu metrics in a database called `kapacitor_example`.
```
gunnaraasen: An InfluxDB query to confirm metrics are in the database would be helpful. Something like:

```
curl -G 'http://localhost:8086/query?db=kapacitor_example' --data-urlencode 'q=SELECT count(value) FROM cpu'
```

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+```sh
+kapacitord config > kapacitor.conf
+```
+
+The configuration is a [toml](https://github.com/toml-lang/toml) file and is very similar to the InfluxDB configuration. 
+That is because any input you can configure for InfluxDB also works for Kapacitor.
+
+There are a few pieces of the configuration that need some attention.
+
+* If your InfluxDB cluster is running on `http://localhost:8086` then you are good to go.
```
gunnaraasen: Is this a requirement for Kapacitor or what's assumed for the examples in this guide? If it's the later, this should say, "all examples assume InfluxDB is running locally at `http://localhost:8086`".

Also, strike the word "cluster" since there's only a single node running for the example.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+    If not then update the configuration with the correct connection settings for InfluxDB in the `influxdb` section.
+
+* In the default configuration file you should see that it is going to store some information in the `~/.kapacitor` directory.
```
gunnaraasen: This will eventually be installation dependent. Data location can be mentioned on an installation page but doesn't really need to be mentioned here.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+    If that directory is alright with you then leave it and Kapacitor will create it on startup if necessary.
+
+* Since Kapacitor will automatically subscribe to any data InfluxDB is receiving lets configure Kapacitor to just subscribe to the
```
gunnaraasen: > Kapacitor will automatically subscribe to any data InfluxDB is receiving lets configure Kapacitor to just subscribe

If it automatically subscribes then why do we need to configure the subscription?

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+    ```
+    [influxdb.subscriptions]
+        kapacitor_example = ["default"]
+    ```
+
+    That configuration tells Kapacitor to only subscribe to the database `kapacitor_example` and retention policy `default`.
+
+* InfluxDB will need to be able to connect to the Kapacitor server so we need to tell Kapacitor its hostname that others will be able to resolve.
+    Check the `hostname` option at the top of the configuration file and make sure that your InfluxDB server can resolve it correctly.
```
gunnaraasen: How do we check whether InfluxDB can correctly resolve the hostname? Will there be a connection failure in the logs?

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+At this point Kapacitor has connected to InfluxDB and configured a [subscription](https://github.com/influxdb/influxdb/blob/master/influxql/INFLUXQL.md#create-subscription)
+for the `kapacitor_example` database. InfluxDB is streaming the data it is receiving from Telegraf to Kapacitor over UDP.
+
+Let's confirm that kapacitord is receiving data from InfluxDB. Run this command to turn on debug logging.
+
+```sh
+kapacitor level debug
+```
+
+You should see a bunch of lines with numbers scrolling by. If you do and the numbers are increasing then we are good to go.
```
gunnaraasen: Might want to give a better explanation of what the numbers mean.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+stream
+    // Select just the cpu_usage_idle measurement from our example database.
+    .from('''"kapacitor_example"."default"."cpu_usage_idle"''')
+    .alert()
+        .crit("value <  70")
+        // Whenever we get an alert write it to a file.
+        .log("./alerts.log")
+```
+
+Put the above script into a file called `cpu_alert.tick` in your working directory.
```
gunnaraasen: This feels like it should be combined with the line above the code block. The code block needs a better introduction so users understand that it represents a stream task.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+Now lets define the `task`. Kapacitor has an HTTP API with which all communcation happens.
```
gunnaraasen: May want to move this paragraph up to where `kapacitor` is first used.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+The binary `kapacitor` calls the API and makes using Kapacitor on the command line easy.
+
+NOTE: If `kapacitord` is not running on `localhost:9092` then pass the `-url` flag to the `kapacitor` commands
```
gunnaraasen: Also move this to where `kapacitor` is first introduced/used.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+That's it, Kapacitor now knows how to trigger our alert, but nothing is going to happen until we enable the task.
+Before we enable the task we should test it first so we do not spam ourselves with alerts.
+
+Record the current data stream for a bit so we can use it to test our task.
+
+```sh
+kapacitor record stream -duration 20s
+```
+
+Now grab that ID that was returned and lets put it in a bash variable for easy use:
```
gunnaraasen: May want to use the raw recording ID here or at least show an example of what an ID should look like.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+OK, we have a snapshot of data recorded from the stream so we can now replay that data to our task.
+The `replay` action replays data only to a specific task this way we can test the task in complete isolation.
+
+```sh
+kapacitor replay -id $rid -name cpu_alert -fast
+```
+
+Notice the `-fast` flag.
+Since we already have the data recorded we can just replay the data as fast as possible instead of waiting for real time to pass.
```
gunnaraasen: What happens if `-fast` is not specified? Is data replayed using the deltas between the original timestamps?

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+Depending on how busy the server was maybe not. Let's modify the task to be really sensitive so that we know the alerts are working.
+Change the `.crit("value < 70")` line in the TICK script to `.crit("value < 100")`. Now every data point that was received during 
+the recording will trigger an alert.
+
+
+Let's replay it again and verify.
+
+```sh
+# Fisrt re-define the task
```
gunnaraasen: Fisrt -> First

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+kapacitor define -name cpu_alert -type stream -tick cpu_alert.tick
+# Now replay it
+kapacitor replay -id $rid -name cpu_alert -fast
+```
+
+Now that we know its working let's change it back to a more reasonable threshold.
```
gunnaraasen: its -> it's

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+
+```sh
+# Fisrt re-define the task
```
gunnaraasen: Fisrt -> First

content/docs/kapacitor/v0.1/introduction/getting_started.md
gunnaraasen: Add a comment that the commands above should be run again to pick up the change back to the original threshold. Then the repeated code blocks below this line can be removed.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+Notice the `-fast` flag.
+Since we already have the data recorded we can just replay the data as fast as possible instead of waiting for real time to pass. 
+Time is measured by the data points that each node receives; therefore the result is identical whether real time passes or not.
+
+
+Check the log, did we get any alerts?
+
+```sh
+cat ./alerts.log
```
gunnaraasen: What does an alert look like?

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+```sh
+while true; do i=0; done
+```
+
+Well that was cool and all but just to get a simple threshold alert there are plenty of ways to do that so why all this
+pipeline TICK script stuff? Well, it can quickly be extended to become much more powerful.
+
+This TICK script will compute the running mean and compare current values to it.
+It will then trigger an alert if the values are more than `3` standard deviations away from the mean.
```
gunnaraasen: 3 doesn't need back ticks

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+stream
+    .from('''"kapacitor_example"."default"."cpu_usage_idle"''')
+    .alert()
+        // Compare values to running mean and standard deviation
+        .crit("sigma(value) > 3")
+        .log("./alerts.log")
+```
+
+Just like that we have a dynamic threshold and if cpu usage drops in the day or spike at night we will get an alert!
+Try it out if you like.  Redefining a running task will not change the running task. Use `reenable` in order to get a
```
gunnaraasen: > Redefining a running task will not change the running task.

Strike this sentence. The sentence after explains that a task needs to be reloaded.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+running task to update based on a new definition.
+
+```sh
+kapacitor reenable cpu_alert
+```
+
+Once you get metrics from all you hosts streaming to Kapacitor you could do something like this:
```
gunnaraasen: you -> your

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+standard deviations away from normal behavior as defined by the historical 95th percentile of cpu usage, within 1 minute!
+Now that is power.
+
+
+
+Trigger Alert from Batch data
+------------------------------
+
+Instead of just processing the data as it streams in you can tell Kapacitor to periodically query
+InfluxDB and then process that data in batch.
```
gunnaraasen: batch -> batches

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+
+
+The batch recording can be replayed in the same way:
+
+```sh
+kapacitor replay -id $rid -name batch_cpu_alert
+```
+
+That's it! Go have fun setting up all kinds of fun and interesting pipelines.
+The rest of the documentation covers the many features and details in using Kapacitor, have a look around.
```
gunnaraasen: Add some links on where to go next.

content/docs/kapacitor/v0.1/introduction/getting_started.md
```diff
@@ -0,0 +1,384 @@
...
+Trigger Alert from Batch data
+------------------------------
+
+Instead of just processing the data as it streams in you can tell Kapacitor to periodically query
+InfluxDB and then process that data in batch.
+While triggering an alert based off cpu usage is more suited for the streaming case you can get the basic idea
+of how `batch` tasks work by following the same use case.
+
+
+This TICK script does the same thing as the earlier stream task but as a batch task.
```
gunnaraasen: Explain batch task a bit more, e.g. it queries data already in influx, better for some types of alerts, etc.

content/docs/kapacitor/v0.1/introduction/index.md
```diff
@@ -0,0 +1,21 @@
...
+
+Kapacitor is a time series data processing engine.
+Kapacitor is the final piece of the TICK stack.
+
+* T - Telegraf -- Data collection
+* I - InfluxDB -- Data storage
+* C - Chronograf -- Data visualization
+* K - Kapacitor -- Data processing
+
+Kapacitor provides a way for you to define a pipeline for processing time series data.
```
gunnaraasen: Put this paragraph before the TICK stack explanation.

content/docs/kapacitor/v0.1/expr/index.md
```diff
@@ -0,0 +1,45 @@
+---
+title: Expression Language
```
gunnaraasen: How about conditional language instead of expression language?

Maybe there are other uses for these expressions, but it sounds like they'll reduce to conditional statements which will trip a trigger when true.

content/docs/kapacitor/v0.1/tick/index.md
```diff
@@ -0,0 +1,123 @@
...
+------
+
+### Strings
+
+Strings are enclosed in double quotes `"` or triple single quotes `'''`.
+There is no escaping in strings, they are treated a literals inside their respective quotes.
+
+### Numbers
+
+Numbers are typed and are either a `float64` or an `int64`. If the number contains a decimal it is considered to be a `float64` otherwise it is an `int64`.
```
gunnaraasen: The InfluxDB line protocol switched from using no decimal to indicate integers to using an explicit `i` a few months ago. Not sure if Kapacitor will run into the same problems as the InfluxDB line protocol, but it could be worth considering a switch to an explicit `i` for integer syntax.

content/docs/kapacitor/v0.1/expr/index.md
```diff
@@ -0,0 +1,45 @@
...
+As soon as a value is processed that is more than `3` standard deviation it evaluates to `true`.
+Now you can use that expression inside of a TICK script to define powerful alerts.
+
+TICK Script:
+
+```javascript
+stream
+    .alert()
+        // use an expression to define when an alert should go critical.
+        .crit("sigma(value) > 3")
```
gunnaraasen: Is there a reason expressions are double quoted? Seems like unnecessary padding unless it's needed by the parser.

nathanielc: Yes, so that's a good question and something I'd like feedback on. 

Right now there are two different parsers, one for TICKscript and  one for the Expression language. 

The expression language is evaluated for each data point while the TICKscript is evaluated once when defining the task. Since they serve different purposes and have very different performance needs I thought keeping them separate was a good idea.

Thoughts?

content/docs/kapacitor/v0.1/tick/index.md
nathanielc: That's a good point. Its pretty common for mathematical expression to infer the type from the decimal point(python, fortran, go etc) but explicit might be better especially since we already do it elsewhere.

I'll think on this.

content/docs/kapacitor/v0.1/expr/index.md
nathanielc: The language can also be used to transform data points:

For example:
```
.apply(expr("newname", "value * 10"))
```

So expression language fits for me.

gunnaraasen: Good point.

On another note, for the example you gave, I would expect the syntax to be something like:

```
.apply("value * 10")
  .rename("newname")
```

nathanielc: I do like that better, I'll create an issue in Kapacitor.

gunnaraasen: Also, it feels like "language" implies that there's more to the syntax than basic mathematical expressions. How about simply using the terminology "expressions" and "conditions".

Then you could say some nodes take expressions, like `.apply()`. Others take conditions, like `.crit()`. Functions, like `sigma()` can be used in expressions (some of which are built in).

content/docs/kapacitor/v0.1/tick/index.md
gunnaraasen: I don't feel strongly either way. I'm not exactly sure how InfluxQL handles int/floats. The only big pluses for using `i` would be consistency with the InfluxDB line protocol and visual clarity for human eyes.

The line protocol adopted `i` because some client libraries would write floats without the decimal, e.g. `1.0` as `1`, causing the type of a field to be cast as an int when a float was desired. A field would then be stuck as a float and InfluxDB would return a type mismatch error.

I don't believe Kapacitor has the same problem with incorrect type casting resulting in errors. There could be problems if a function only, like sigma, expects a float and receives an int, but that could be handled in different ways.

content/docs/kapacitor/v0.1/expr/index.md
gunnaraasen: I'm partial to fewer keystrokes where possible. I think good documentation would clear up any confusion that expressions are evaluated for each point. I'll defer to @pauldix's opinion on this one.

nathanielc:
@gunnaraasen how does it look now? There are several fixes the the TICKscript node docs as well as a 'Use Cases' section that for now has one more complex use case of Kapacitor. I will be adding several more soon.

pauldix:
Just skimmed, looks good! +1

nathanielc:
@pauldix Thanks, talked with @gunnaraasen  he is on board, merging...

Srungaram:
I am new to InfluxDB+ Kapacitor. I am not using Telegraf to collect the data as I have my own collector app.
My Question is Where do I put the tick script?
It is mentioned in the documentation to place it in working directory. what does this mean?
I have tried placing it in kapacitor-0.1.0.2_linux_amd64/tmp/package-dir-staging20151117-3986-t58jwa/usr/bin folder but when I run ./kapacitor record stream -name cpu_alert -duration 20s
it is saying "no tasks bucket". Please help

nathanielc:
@Srungaram  You can put the tick script anywhere. When you run the command: `kapacitor define ` you just give it the path to your tick script and it uploads it to the server. 

The example in the docs shows:

```sh
kapacitor define \
    -name cpu_alert \
    -type stream \
    -tick cpu_alert.tick \
    -dbrp kapacitor_example.default
```
Assuming the file `cpu_alert.tick` exists in your current directory.

Based on the error "no tasks bucket" sound like the db may have been deleted try restarting the kapacitord daemon and let it reinitialize. 

FYI Kapacitor has not been officially released yet but we are glad you have interest. Feel free to contact me directly if you need more help getting it working since you won't be able to file issues on the github project yet. I have contact info on my GH profile.

PR #422, opened 2015-10-21
edit glossary.md and key_concepts.md

rkuchan:
Edits to `glossary.md`
---

* Alphabetize it!
* Link to docs where possible
* Add 'Related entries' to entries that reference other Glossary terms 
* Add entries for identifier, hinted handoff, query

In general, I tried to make the glossary colder - I removed examples and some longer explanations as I tried to incorporate that kind of discussion in `key_concepts.md`.

Edits to `key_concepts.md`
---

I **tried** to make this more of a general, gradual introduction to InfluxDB's data structure and concepts. It attempts to tell the InfluxDB story through simple example data. 👀s note where I made possibly not-OK simplifications or where I'd like some feedback - but in general, feedback on the whole thing would be great!

To do: 
* Add wal, shard, shard group, storage engines (tsm1,b1,bz1) to `glossary.md`

### DIFF
--- a/content/docs/v0.9/concepts/glossary.md
+++ b/content/docs/v0.9/concepts/glossary.md
@@ -8,15 +8,22 @@ An InfluxQL function that returns an aggregated value across a set of points. Se
 Related entries: [function](../concepts/glossary.html#function), [selector](../concepts/glossary.html#selector), [transformation](../concepts/glossary.html#transformation)
 
 ## cluster
-A collection of servers running InfluxDB nodes. See [Clustering](../guides/clustering.html) for how to set up an InfluxDB cluster.
+A collection of servers running InfluxDB nodes. All nodes in a cluster have the same users, databases, retention policies, and continuous queries. See [Clustering](../guides/clustering.html) for how to set up an InfluxDB cluster.
 
 Related entries: [node](../concepts/glossary.html#node), [server](../concepts/glossary.html#server)
 	
 ## continuous query (CQ)
-An InfluxQL query that runs automatically and periodically within a database. See [Continuous Queries](../query_language/continuous_queries.html).
+An InfluxQL query that runs automatically and periodically within a database. Continuous queries require a function in the `SELECT` clause and must include a `GROUP BY time()` clause. See [Continuous Queries](../query_language/continuous_queries.html).  
+
+Related entries: [function](../concepts/glossary.html#function)
+
+## coordinator node
+The node that receives write and query requests for the cluster.  
+
+Related entries: [cluster](../concepts/glossary.html#cluster), [hinted handoff](../concepts/glossary.html#hinted-handoff), [node](../concepts/glossary.html#node)
 	
 ## database
-A logical container for users, retention policies, continuous queries, and timeseries data.
+A logical container for users, retention policies, continuous queries, and time series data.
 
 Related entries: [continuous query](../concepts/glossary.html#continuous-query-cq), [retention policy](../concepts/glossary.html#retention-policy-rp), [user](../concepts/glossary.html#user)
 
@@ -26,16 +33,14 @@ The attribute of the retention policy that determines how long InfluxDB stores d
 Related entries: [replication factor](../concepts/glossary.html#replication-factor), [retention policy](../concepts/glossary.html#retention-policy-rp) 
 
 ## field
-The key-value pair in InfluxDB's data structure that records metadata and the actual data value. Fields are required in InfluxDB's data structure and they are not indexed - queries on fields scan all points that match the specified time range and, as a result, are not performant.
+The key-value pair in InfluxDB's data structure that records metadata and the actual data value. Fields are required in InfluxDB's data structure and they are not indexed - queries on field values scan all points that match the specified time range and, as a result, are not performant relative to tags.
 
 *Query tip:* Compare fields to tags; tags are indexed.
 
 Related entries: [field key](../concepts/glossary.html#field-key), [field set](../concepts/glossary.html#field-set), [field value](../concepts/glossary.html#field-value), [tag](../concepts/glossary.html#tag)
 	
 ## field key
-The key part of the key-value pair that makes up a field. Field keys are strings and they store metadata. Field keys are not indexed - queries on field keys scan all points that match the specified time range and, as a result, are not performant.
-
-*Query tip:* Compare field keys to tag keys; tag keys are indexed.
+The key part of the key-value pair that makes up a field. Field keys are strings and they store metadata. 
 
 Related entries: [field](../concepts/glossary.html#field), [field set](../concepts/glossary.html#field-set), [field value](../concepts/glossary.html#field-value), [tag key](../concepts/glossary.html#tag-key)
 	
@@ -47,7 +52,7 @@ Related entries: [field](../concepts/glossary.html#field), [field key](../concep
 ## field value  
 The value part of the key-value pair that makes up a field. Field values are the actual data; they can be strings, floats, integers, or booleans. A field value is always associated with a timestamp.
 
- Field values are not indexed - queries on field values scan all points that match the specified time range and, as a result, are not performant.
+Field values are not indexed - queries on field values scan all points that match the specified time range and, as a result, are not performant.
 
 *Query tip:* Compare field values to tag values; tag values are indexed.
 
@@ -59,9 +64,9 @@ InfluxQL aggregations, selectors, and transformations. See [InfluxQL Functions](
 Related entries: [aggregation](../concepts/glossary.html#aggregation), [selector](../concepts/glossary.html#selector), [transformation](../concepts/glossary.html#transformation)  
 
 ## hinted handoff
-A durable write queue for any writes missed by a server. Nodes temporarily store queued data when one node of a cluster is down for a short period of time. 
+A durable queue of data destined for a server which was unavailable at the time the data was received. Coordinating nodes temporarily store queued data when a target node for a write is down for a short period of time.
 
-Related entries: [cluster](../concepts/glossary.html#cluster), [node](../concepts/glossary.html#node)
+Related entries: [cluster](../concepts/glossary.html#cluster), [node](../concepts/glossary.html#node), [server](../concepts/glossary.html#server)
 
 ## identifier
 Tokens which refer to database names, retention policy names, user names, measurement names, tag keys, and field keys. See [Query Language Specification](../query_language/spec.html#identifiers).
@@ -69,31 +74,31 @@ Tokens which refer to database names, retention policy names, user names, measur
 Related entries: [database](../concepts/glossary.html#database), [field key](../concepts/glossary.html#field-key), [measurement](../concepts/glossary.html#measurement), [retention policy](../concepts/glossary.html#retention-policy-rp), [tag key](../concepts/glossary.html#tag-key), [user](../concepts/glossary.html#user)
 
 ## measurement  
-The part of InfluxDB's structure that identifies the unit of the data that are stored in the associated fields. Measurements are strings.
+The part of InfluxDB's structure that describes the data stored in the associated fields. Measurements are strings.
 
 Related entries: [field](../concepts/glossary.html#field), [series](../concepts/glossary.html#series)
 
 ## node
-A logical concept in the cluster setup. A single nodes belongs to a server.
+An independent `influxd` process.
 
 Related entries: [cluster](../concepts/glossary.html#cluster), [server](../concepts/glossary.html#server)
 
 ## point   
-The part of InfluxDB's data structure that consists of the field set in the same series with the same timestamp.
+The part of InfluxDB's data structure that consists of a single collection of fields in a series. Each point is uniquely identified by its series and timestamp.
 
-You cannot store more than one point in the same series with the same timestamp. InfluxDB silently overwrites the old field set with the new field set when you write a new field set to the same series with the same timestamp.
+You cannot store more than one point with the same timestamp in the same series. Instead, when you write a new point to the same series with the same timestamp as an existing point in that series, InfluxDB silently overwrites the old field set with the new field set.
 
 Related entries: [field set](../concepts/glossary.html#field-set), [series](../concepts/glossary.html#series), [timestamp](../concepts/glossary.html#timestamp)
 
 ## query
-An operation that interacts with data in InfluxDB. See [Data Exploration](../query_language/data_exploration.html), [Schema Exploration](../query_language/schema_exploration.html), [Database Management](../query_language/database_management.html).
+An operation that retrieves data from InfluxDB. See [Data Exploration](../query_language/data_exploration.html), [Schema Exploration](../query_language/schema_exploration.html), [Database Management](../query_language/database_management.html).
 	
 ## replication factor  
 The attribute of the retention policy that determines how many copies of the data are stored in the cluster. InfluxDB replicates data across `N` data nodes, where `N` is the replication factor.
 
 To maintain data availability for queries, the replication factor should be less than or equal to the number of data nodes in the cluster:
 
-* Data are available when the replication factor is greater than the number of *unavailable* data nodes. 
+* Data are fully available when the replication factor is greater than the number of *unavailable* data nodes. 
 * Data may be unavailable when the replication factor is less than the number of *unavailable* data nodes.
 
 Note that there are no query performance benefits from replication. Replication is for ensuring data availability when a data node or nodes are unavailable. See [Database Management](../query_language/database_management.html#create-retention-policies-with-create-retention-policy) for how to set the replication factor.
@@ -103,7 +108,7 @@ Related entries: [cluster](../concepts/glossary.html#cluster), [duration](../con
 ## retention policy (RP)
 The part of InfluxDB's data structure that describes for how long InfluxDB keeps data (duration) and how many copies of those data are stored in the cluster (replication factor). RPs are unique per database and along with the measurement and tag set define a series.
 
-By default, InfluxDB automatically creates a retention policy called `default` with an infinite duration and a replication factor of one. See [Database Management](../query_language/database_management.html#retention-policy-management) for retention policy management.
+When you create a database, InfluxDB automatically creates a retention policy called `default` with an infinite duration and a replication factor set to the number of nodes in the cluster. See [Database Management](../query_language/database_management.html#retention-policy-management) for retention policy management.
 	
 Related entries: [duration](../concepts/glossary.html#duration), [measurement](../concepts/glossary.html#measurement), [replication factor](../concepts/glossary.html#replication-factor), [series](../concepts/glossary.html#series), [tag set](../concepts/glossary.html#tag-set)
 
@@ -171,15 +176,15 @@ There are two kinds of users in InfluxDB:
 * *Admin users* have `READ` and `WRITE` access to all databases and full access to administrative queries and user management commands.
 * *Non-admin users* have `READ`, `WRITE`, or `ALL` (both `READ` and `WRITE`) access per database. 
 
-When authentication is enabled, InfluxDB only executes HTTP request that are sent with a valid username and password. See [Authentication and Authorization](../administration/authentication_and_authorization.html).
+When authentication is enabled, InfluxDB only executes HTTP requests that are sent with a valid username and password. See [Authentication and Authorization](../administration/authentication_and_authorization.html).
 
 <!--
-## Column
+## wal
 
-## Batch
+## shard
 
-## Shard
+## shard group
 
-## Shard File
+## storage engines (tsm1, b1, bz1)
 
 -->
--- a/content/docs/v0.9/concepts/glossary.md
+++ b/content/docs/v0.9/concepts/glossary.md
@@ -2,216 +2,184 @@
 title: Glossary of Terms
 ---
 
-## Series
-A **series** is a collection of data points along a timeline that share a common key, expressed as a **measurement** and **tag set** pairing, grouped under a **retention policy**.
+## aggregation
+An InfluxQL function that returns an aggregated value across a set of points. See [InfluxQL Functions](../query_language/functions.html#aggregations) for a complete list of the available and upcoming aggregations.
 
-```
-<measurement>,<tag set>
-```
+Related entries: [function](../concepts/glossary.html#function), [selector](../concepts/glossary.html#selector), [transformation](../concepts/glossary.html#transformation)
 
-> **Note:** The [field set](/docs/v0.9/concepts/glossary.html#field) is not part of the series identification.
+## cluster
+A collection of servers running InfluxDB nodes. See [Clustering](../guides/clustering.html) for how to set up an InfluxDB cluster.
 
-## Measurement
-A **measurement** is the name of the value that is recorded in the **series**, and can be shared amongst many **series**.
+Related entries: [node](../concepts/glossary.html#node), [server](../concepts/glossary.html#server)
+	
+## continuous query (CQ)
+An InfluxQL query that runs automatically and periodically within a database. See [Continuous Queries](../query_language/continuous_queries.html).
+	
+## database
+A logical container for users, retention policies, continuous queries, and timeseries data.
 
-The **measurement** should describe the metric(s) recorded in the associated **series** and should not explicitly include information about the source of or the metadata associated with what is being measured. Metadata should be recorded in **tags**.
+Related entries: [continuous query](../concepts/glossary.html#continuous-query-cq), [retention policy](../concepts/glossary.html#retention-policy-rp), [user](../concepts/glossary.html#user)
 
-All **series** under a given **measurement** have the same **field keys** and differ only in their **tag set**.
+## duration
+The attribute of the retention policy that determines how long InfluxDB stores data. Data older than the duration are automatically dropped from the database. See [Database Management](../query_language/database_management.html#create-retention-policies-with-create-retention-policy) for how to set duration.
+	
+Related entries: [replication factor](../concepts/glossary.html#replication-factor), [retention policy](../concepts/glossary.html#retention-policy-rp) 
 
-The **measurement** is what appears in the SELECT clause of query statements, and is thus similar to a table name in traditional SQL databases.
+## field
+The key-value pair in InfluxDB's data structure that records metadata and the actual data value. Fields are required in InfluxDB's data structure and they are not indexed - queries on fields scan all points that match the specified time range and, as a result, are not performant.
 
-#### Example:
+*Query tip:* Compare fields to tags; tags are indexed.
 
-The metrics `cpu_load`, `energy_generated`, and `bikes_present` would be reasonable **measurements** for a server, a photovoltaic array, and a bike-sharing station, respectively.
+Related entries: [field key](../concepts/glossary.html#field-key), [field set](../concepts/glossary.html#field-set), [field value](../concepts/glossary.html#field-value), [tag](../concepts/glossary.html#tag)
+	
+## field key
+The key part of the key-value pair that makes up a field. Field keys are strings and they store metadata. Field keys are not indexed - queries on field keys scan all points that match the specified time range and, as a result, are not performant.
 
-The `cpu_load` measurement would likely have tags for hostname, server region, server function, and OS. The `energy_generated` measurement might have tags for pv_installation_id, outside_temperature, and weather. The `bikes_present` measurement might only have a tag for station_id.
+*Query tip:* Compare field keys to tag keys; tag keys are indexed.
 
-## Point
-A **point** is a single collection of **fields** in a **series**. Each **point** is uniquely identified by its **series** and **timestamp**.
+Related entries: [field](../concepts/glossary.html#field), [field set](../concepts/glossary.html#field-set), [field value](../concepts/glossary.html#field-value), [tag key](../concepts/glossary.html#tag-key)
+	
+## field set
+The collection of field keys and field values on a point.
 
-It is not possible to store two **points** in the same **series** with the same timestamp. The most recently written version of the **point** will silently overwrite the previous **field set** with its **field set**.
+Related entries: [field](../concepts/glossary.html#field), [field key](../concepts/glossary.html#field-key), [field value](../concepts/glossary.html#field-value), [point](../concepts/glossary.html#point)  
+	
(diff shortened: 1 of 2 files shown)


content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
 
-## Measurement
-A **measurement** is the name of the value that is recorded in the **series**, and can be shared amongst many **series**.
+Related entries: [node](../concepts/glossary.html#node), [server](../concepts/glossary.html#server)
+	
+## continuous query (CQ)
+An InfluxQL query that runs automatically and periodically within a database. See [Continuous Queries](../query_language/continuous_queries.html).
+	
+## database
+A logical container for users, retention policies, continuous queries, and timeseries data.
```
otoolep: "time series" is two words.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
 
-All **series** under a given **measurement** have the same **field keys** and differ only in their **tag set**.
+## duration
+The attribute of the retention policy that determines how long InfluxDB stores data. Data older than the duration are automatically dropped from the database. See [Database Management](../query_language/database_management.html#create-retention-policies-with-create-retention-policy) for how to set duration.
+	
+Related entries: [replication factor](../concepts/glossary.html#replication-factor), [retention policy](../concepts/glossary.html#retention-policy-rp) 
 
-The **measurement** is what appears in the SELECT clause of query statements, and is thus similar to a table name in traditional SQL databases.
+## field
+The key-value pair in InfluxDB's data structure that records metadata and the actual data value. Fields are required in InfluxDB's data structure and they are not indexed - queries on fields scan all points that match the specified time range and, as a result, are not performant.
```
otoolep: We should be careful here. InfluxDB performs well, it's all relative. Queries that involve scanning points are not as performant as queries that only have to look at data in the in-memory index i.e. tags. So it's not perfomant *relative to tags*.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
 
-#### Example:
+*Query tip:* Compare fields to tags; tags are indexed.
 
-The metrics `cpu_load`, `energy_generated`, and `bikes_present` would be reasonable **measurements** for a server, a photovoltaic array, and a bike-sharing station, respectively.
+Related entries: [field key](../concepts/glossary.html#field-key), [field set](../concepts/glossary.html#field-set), [field value](../concepts/glossary.html#field-value), [tag](../concepts/glossary.html#tag)
+	
+## field key
+The key part of the key-value pair that makes up a field. Field keys are strings and they store metadata. Field keys are not indexed - queries on field keys scan all points that match the specified time range and, as a result, are not performant.
```
otoolep: Ditto, I think we should phrase this differently

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
+InfluxQL aggregations, selectors, and transformations. See [InfluxQL Functions](../query_language/functions.html) for a complete list of InfluxQL functions.
 
-#### Example
-The following are all valid **tags**: `hostname=server01`, `station_id=84`, and `weather='partial sun'`.
+Related entries: [aggregation](../concepts/glossary.html#aggregation), [selector](../concepts/glossary.html#selector), [transformation](../concepts/glossary.html#transformation)  
 
-## Tag Key
-The **tag key** is the key part of the key-value pair that makes up a **tag** and is always stored as a string.
+## hinted handoff
+A durable write queue for any writes missed by a server. Nodes temporarily store queued data when one node of a cluster is down for a short period of time.
```
otoolep: "missed by a server" sounds like the target server was up and did something wrong. I think this might be better phrased as:

"A durable queue of data destined for a server which was unavailable at the time the data was received. Co-ordinating nodes temporarily store queued data when a target node for a write is down for a short period of time".

This reminds me -- we should add the concept "co-ordinator node". This is a well-known term which means the node that receives write and query requests for the cluster. In this case it is the  "co-ordinator node" that stores the HH data.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
+## measurement  
+The part of InfluxDB's structure that identifies the unit of the data that are stored in the associated fields. Measurements are strings.
 
-If you write multiple **tag values** for the same **tag key** on a given **point**, one of the **tag values** will be assigned to the **tag key** and all other **tag values** will be silently dropped.
+Related entries: [field](../concepts/glossary.html#field), [series](../concepts/glossary.html#series)
 
-## Tag Set
-The **tag set** is the combination of all **tags** on the **point**, including all **tag keys** and their corresponding **tag values**.
+## node
+A logical concept in the cluster setup. A single nodes belongs to a server.
```
otoolep: Not sure about the second sentence. What are you trying to say?

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
-The total number of **series** under a given **measurement** is thus identical to the number of distinct **tag sets** applied to **points** within that **measurement**.
+You cannot store more than one point in the same series with the same timestamp. InfluxDB silently overwrites the old field set with the new field set when you write a new field set to the same series with the same timestamp.
 
-## Field
-A **field** is a key-value pair that records an actual metric for a given **point**.
+Related entries: [field set](../concepts/glossary.html#field-set), [series](../concepts/glossary.html#series), [timestamp](../concepts/glossary.html#timestamp)
 
-A **field** is made up of the **field key** and the **field value**.
+## query
+An operation that interacts with data in InfluxDB. See [Data Exploration](../query_language/data_exploration.html), [Schema Exploration](../query_language/schema_exploration.html), [Database Management](../query_language/database_management.html).
```
otoolep: interacts -> retrieves?

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
 title: Key Concepts
 ---
 
-To understand InfluxDB, it is necessary to understand some key concepts. These concepts are outlined below.
+Before diving into InfluxDB it's good to get acquainted with some of the key concepts of the database. This document provides a gentle introduction to those concepts and common InfluxDB terminology. We've provided a list below of all the terms we'll cover, but we recommend reading this document chronologically to gain a more general understanding of your favorite timeseries database.
```
beckettsean: "understanding of **your** favorite timeseries database."

I think this is a bit too presumptuous here. If I were reading the docs as part of a product comparison I would find this a bit too cocky or familiar. How about

"understanding of **our** favorite timeseries database. "

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+[tag set](../concepts/key_concepts.html#tag-set)&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+[tag value](../concepts/key_concepts.html#tag-value)&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;
+[timestamp](../concepts/key_concepts.html#timestamp)  
 
-## Time-Series Data
+Check out the [Glossary](../concepts/glossary.html) if you prefer the cold, hard facts.
 
-InfluxDB uses particular terms to describe the various components of time-series data, and the techniques used to categorize that data.
+### Sample data
+The next section references the data printed out below. The data are fictional, but represent a believable setup in InfluxDB. Assume that the data live in a database called, rather uninterestingly, `my_database` and are subject to the `default` retention policy (more on databases and retention policies to come).
```
beckettsean: I like "rather uninterestingly" but it's gonna be tough on ESL folks. Keep, change, or remove, up to you. I'm fine with all three.

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
 
-## Time-Series Data
+Check out the [Glossary](../concepts/glossary.html) if you prefer the cold, hard facts.
 
-InfluxDB uses particular terms to describe the various components of time-series data, and the techniques used to categorize that data.
+### Sample data
+The next section references the data printed out below. The data are fictional, but represent a believable setup in InfluxDB. Assume that the data live in a database called, rather uninterestingly, `my_database` and are subject to the `default` retention policy (more on databases and retention policies to come). 
 
-Every data point within InfluxDB has a **timestamp**. When a point is written to InfluxDB it is either accompanied by a timestamp or, if it has no associated timestamp, InfluxDB uses its local clock to assign a timestamp to the point. It is important to note that InfluxDB does not differentiate between points that came with a timestamp versus points that were assigned a timestamp at reception time. 
+*Hint:* Hover over the links to briefly get acquainted with InfluxDB terminology and the layout. It's probably best not to click on them unless you're already somewhat familiar with InfluxDB.
```
beckettsean: "Hover over the links for tooltips to get acquainted with InfluxDB terminology and the layout. It's probably best not to follow the links until you are somewhat familiar with InfluxDB."

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
 
-A **series** is defined as a combination of a _measurement_ and set of _tag_ key-values. Combined with _fields_ (columns) and the fields' _values_, these make up series _data_.
+👀 I like the idea of being able to see the location of the measurement, tags, fields, etc. in a standard data output. I couldn't figure out a better way to do this in markdown - is my method kind of annoying or do you think it offers some useful information? 👀
```
beckettsean: I really like it. Tooltips are a known pattern for folks and doing it in the column output gives people familiarity with the CLI.

Maybe we need to put the JSON return in parallel here? Of if not here, perhaps we have a new page that compares the outputs? Or perhaps we put some tool-tipped JSON return on a Query page?

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+2015-08-18T06:12:00Z&nbsp;&nbsp;&nbsp;7	&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;22	&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;perpetua  
 
-InfluxDB also allows you to associate **tags** with measurements. Tags are arbitrary key-value pairs associated with a single time-series data point. Series data are indexed by tags, allowing efficient and fast look-up of series data that match a given set of tags. Finally a **field** is the part of a time-series data point that is not indexed by the system.
+### Discussion
+Now that you've seen some sample data in InfluxDB this section covers [what it all means](https://www.youtube.com/watch?v=I5cYgRnfFDA).
```
beckettsean: Not sure what I expected when I clicked that link, but it wasn't that YouTube clip. It's cute but maybe a little forced.

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+3    28
+2    11
+1    10
+8    23
+7    22
+```
+
+The key part of the key-value pair that makes up a field is called the <a name="field-key"></a> **field key**. Field keys are strings and they store metadata; the field key `butterflies` tells us that the field values `12`-`7` refer to butterflies and the field key `honeybees` tells us that the field values `23`-`22` refer to, well, honeybees. 
+
+Fields are a required piece of InfluxDB's data structure - you cannot have data in InfluxDB without fields. It's also important to note that fields are not indexed - [queries](../concepts/glossary.html#query) on fields scan all points that match the specified time range in the query and, as a result, are not performant. Because of this, the metadata stored in the field keys should not contain commonly-queried metadata. We'll cover where to store commonly-queried metadata later.
```
beckettsean: [queries](../concepts/glossary.html#query) with fields on the WHERE clause must scan all points that match the other conditions in the query and, as a result, are not performant.

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+* `butterflies = 12   honeybees = 23`
+* `butterflies = 1    honeybees = 30`
+* `butterflies = 11   honeybees = 28`
+* `butterflies = 3    honeybees = 28`
+* `butterflies = 2    honeybees = 11`
+* `butterflies = 1    honeybees = 10`
+* `butterflies = 8    honeybees = 23`
+* `butterflies = 7    honeybees = 22`
+
+👀 Is this an OK simplification? To get the formal definition of a field set they'd need to know what a point is and I don't want to get into series until they know what measurements, tag sets, and retention policies are. 👀
```
beckettsean: Seems good to me so far. We won't really know until we put this in front of a true InfluxDB noob.

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+
+As we mentioned earlier, commonly queried metadata don't belong in fields because fields aren't indexed. Tags, on the other hand, are indexed - this means that queries on tags are fast and that tags are ideal for storing commonly-queried metadata.  A tag is made up of a <a name="tag-key"></a>**tag key** and a <a name="tag-value"></a>**tag value**. Both tag keys and tag values are stored as strings. Tags are also optional; you don't need to have tags in your data, but it's generally a good idea to make use of them.
```
beckettsean: I'd edit "As we mentioned earlier, commonly queried metadata don't belong in fields because fields aren't indexed." to 

"Metadata commonly used as filters in queries doesn't belong in fields because fields aren't indexed."

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+
+The tag keys in the sample data are `location` and `scientist`. The tag key `location` has two tag values: `1` and `2`. The tag key `scientist` also has two tag values: `langstroth` and `perpetua`.
+
+In the data above, the <a name="tag-set"></a>**tag set** is the different combinations of all the tag key-value pairs that have the same timestamp. The four tag sets in the sample data are:
```
beckettsean: I'd strike "that have the same timestamp". It confuses me and I know what you're trying to say. I don't think we need to be that worried about absolute correctness, yet.

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+2015-08-18T06:06:00Z&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;23&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;honeybees  
+2015-08-18T06:12:00Z&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;7&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;butterflies  
+2015-08-18T06:12:00Z&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;22&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;2&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;&nbsp;honeybees  
+
+> Now your queries on the tag values `honeybees` and `butterflies` are performant.
```
beckettsean: I really like this approach, and it will help tremendously with schema design questions.

I'd love this to be copied to a new schema design page. On that page provide a set of example queries, each being performant in one schema and non-performant in the other. That way people can more easily reason about the schema they need.

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+
+The <a name=measurement></a>**measurement** is the unit of the data that are stored in the associated fields. Measurements are strings, and, for any SQL users out there, a measurement is similar to a table. The only measurement in the sample data is `count`. `count` tells us that field values are the number of `butterflies` and `honeybees` - not some sort of happiness index.
```
beckettsean: "is the unit of the data that are stored in the associated fields."

"unit" feels weird here. Can easily be confused with the units of the metric, like `m/s` or `kg` or `kilobytes`.

maybe

"The <a name=measurement></a>**measurement** is the description of the metrics that are stored in the associated fields."

beckettsean: "a measurement is conceptually similar to a table."

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+*Hint:* Hover over the links to briefly get acquainted with InfluxDB terminology and the layout. It's probably best not to click on them unless you're already somewhat familiar with InfluxDB.
 
-A **series** is defined as a combination of a _measurement_ and set of _tag_ key-values. Combined with _fields_ (columns) and the fields' _values_, these make up series _data_.
+👀 I like the idea of being able to see the location of the measurement, tags, fields, etc. in a standard data output. I couldn't figure out a better way to do this in markdown - is my method kind of annoying or do you think it offers some useful information? 👀
 
-A **measurement** is the value being recorded in the series. For example `cpu_load` or `sensor_temperature`.
+name: [count](../concepts/key_concepts.html#measurement "Measurement")
```
beckettsean: I think we should rename `count` to `tally` or `census` to reduce any confusion with the SQL or InfluxQL `COUNT()` function.

content/docs/v0.9/concepts/key_concepts.md
beckettsean: "The name `count` tells us that field values record the number of `butterflies` and `honeybees` - not their size or direction or some sort of happiness index."

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+
+> Now your queries on the tag values `honeybees` and `butterflies` are performant.
+
+The <a name=measurement></a>**measurement** is the unit of the data that are stored in the associated fields. Measurements are strings, and, for any SQL users out there, a measurement is similar to a table. The only measurement in the sample data is `count`. `count` tells us that field values are the number of `butterflies` and `honeybees` - not some sort of happiness index.
+
+A single measurement can belong to different retention policies. A <a name="retention-policy"></a>**retention policy** describes how long a InfluxDB keeps data (`DURATION`) and how many copies of those data are stored in the cluster (`REPLICATION`). If you're interested in reading more about retention policies, check out [Database Management](../query_language/database_management.html#retention-policy-management).
```
beckettsean: "how long *a* InfluxDB keeps" should be "how long InfluxDB keeps"

content/docs/v0.9/concepts/key_concepts.md
```diff
@@ -2,25 +2,141 @@
...
+| Arbitrary series number  |  Measurement | Tag set  |  Retention policy |
+|---|---|---|---|
+| series 1  | `count` | `location = 1`,`scientist = langstroth`  | `default` |
+| series 2 | `count`  |  `location = 2`,`scientist = langstroth` |  `default` |
+| series 3  | `count` | `location = 1`,`scientist = perpetua`  | `default` |
+| series 4 | `count`  |  `location = 2`,`scientist = perpetua` |  `default` |
+
+Understanding the concept of a series is essential when designing your schema and when working with your data in InfluxDB.
+
+Finally, a <a name="point"></a>**point** is the field set in the same series with the same timestamp. So, for the series where the measurement is `count`; the tag set is `location = 1`,`scientist = perpetua`; and the retention policy is `default`, and for the timestamp `2015-08-18T00:00:00Z`, the point is:
```
beckettsean: I think the point includes the series metadata, too. The tags are included at write time, they just aren't stored in the same place. The field set doesn't define the point, either, only the measurement, tag set, and timestamp can identify a point.

I'm not sure how to reword this, maybe we should brainstorm.

beckettsean:
lots of comments on key concepts but very huge improvement! 

moving onto glossary now

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
+An InfluxQL function that returns an aggregated value across a set of points. See [InfluxQL Functions](../query_language/functions.html#aggregations) for a complete list of the available and upcoming aggregations.
 
-```
-<measurement>,<tag set>
-```
+Related entries: [function](../concepts/glossary.html#function), [selector](../concepts/glossary.html#selector), [transformation](../concepts/glossary.html#transformation)
 
-> **Note:** The [field set](/docs/v0.9/concepts/glossary.html#field) is not part of the series identification.
+## cluster
+A collection of servers running InfluxDB nodes. See [Clustering](../guides/clustering.html) for how to set up an InfluxDB cluster.
```
beckettsean: A collection of servers running InfluxDB nodes. All nodes in a cluster have the same users, databases, retention policies, and continuous queries. See [Clustering](../guides/clustering.html) for how to set up an InfluxDB cluster.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
 
-## Measurement
-A **measurement** is the name of the value that is recorded in the **series**, and can be shared amongst many **series**.
+Related entries: [node](../concepts/glossary.html#node), [server](../concepts/glossary.html#server)
+	
+## continuous query (CQ)
```
beckettsean: Maybe include

Continuous queries must use only functions in the `SELECT` clause and include a `GROUP BY time()`.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
-A **tag value** is the value part of the key-value pair that makes up a **tag** and is always stored as a string.
+## identifier
+Tokens which refer to database names, retention policy names, user names, measurement names, tag keys, and field keys. See [Query Language Specification](../query_language/spec.html#identifiers).
 
-**Tag values** must be unique per **tag key**.
+Related entries: [database](../concepts/glossary.html#database), [field key](../concepts/glossary.html#field-key), [measurement](../concepts/glossary.html#measurement), [retention policy](../concepts/glossary.html#retention-policy-rp), [tag key](../concepts/glossary.html#tag-key), [user](../concepts/glossary.html#user)
 
-InfluxDB will not allow assignment of `weather='partly sunny'` and `weather='partly cloudy'` to the same **point**.
+## measurement  
+The part of InfluxDB's structure that identifies the unit of the data that are stored in the associated fields. Measurements are strings.
```
beckettsean: I have the same reservations with "unit" here that I do on Key Concepts.

content/docs/v0.9/concepts/glossary.md
beckettsean: A node is an independent `influxd` process. I'm not sure what else we want or need to say here. Let's talk about what you wanted to communicate, because I think both @otoolep and I missed it.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
-The **tag set** is the combination of all **tags** on the **point**, including all **tag keys** and their corresponding **tag values**.
+## node
+A logical concept in the cluster setup. A single nodes belongs to a server.
 
-The **tag set** and **measurement** fully describe a **series**.
+Related entries: [cluster](../concepts/glossary.html#cluster), [server](../concepts/glossary.html#server)
 
-The **tag set**, **measurement**, and **timestamp** fully describe a **point**.
+## point   
+The part of InfluxDB's data structure that consists of the field set in the same series with the same timestamp.
```
beckettsean: This sentence is awkward for me, similarly to the Key Concepts definition. I think we need to rework it.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
 
-The total number of **series** under a given **measurement** is thus identical to the number of distinct **tag sets** applied to **points** within that **measurement**.
+You cannot store more than one point in the same series with the same timestamp. InfluxDB silently overwrites the old field set with the new field set when you write a new field set to the same series with the same timestamp.
```
beckettsean: Let's reverse the order, I think it makes it clearer.

You cannot store more than one point with the same timestamp in the same series. Instead, when you write a new point to the same series with the same timestamp as an existing point in that series, InfluxDB silently overwrites the old field set with the new field set.

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
+An operation that interacts with data in InfluxDB. See [Data Exploration](../query_language/data_exploration.html), [Schema Exploration](../query_language/schema_exploration.html), [Database Management](../query_language/database_management.html).
+	
+## replication factor  
+The attribute of the retention policy that determines how many copies of the data are stored in the cluster. InfluxDB replicates data across `N` data nodes, where `N` is the replication factor.
 
-**Fields** are not indexed, meaning queries on **field keys** or **field values** must scan all points matching the time range of the query. Queries on **fields** are thus not performant, but are assumed to be infrequent for typical use cases.
+To maintain data availability for queries, the replication factor should be less than or equal to the number of data nodes in the cluster:
 
-If you find yourself querying **fields** consider whether the key-value pair is better stored as a **tag**.
+* Data are available when the replication factor is greater than the number of *unavailable* data nodes.
```
beckettsean: Data are fully available...

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
+	
+Related entries: [cluster](../concepts/glossary.html#cluster), [duration](../concepts/glossary.html#duration), [node](../concepts/glossary.html#node), [retention policy](../concepts/glossary.html#retention-policy-rp)
 
-Every point must have at least one **Field** or it will be rejected as invalid.
+## retention policy (RP)
+The part of InfluxDB's data structure that describes for how long InfluxDB keeps data (duration) and how many copies of those data are stored in the cluster (replication factor). RPs are unique per database and along with the measurement and tag set define a series.
 
-#### Example
+By default, InfluxDB automatically creates a retention policy called `default` with an infinite duration and a replication factor of one. See [Database Management](../query_language/database_management.html#retention-policy-management) for retention policy management.
```
beckettsean: By default, InfluxDB automatically...

to

When a database is created, InfluxDB automatically...

I really don't need any more `default`s in this section than necessary ;)

content/docs/v0.9/concepts/glossary.md
```diff
@@ -2,216 +2,184 @@
...
-
-## User
-A **user** is required when authentication is enabled for the database.
-
-## Node
-A **node** is a logical concept in the clustering setup. **Nodes** belong to **servers** (one node per server.)
-
-## Server
-A **server** is a machine, virtual or physical, running InfluxDB. There should only be one InfluxDB process per **server**.
+When authentication is enabled, InfluxDB only executes HTTP request that are sent with a valid username and password. See [Authentication and Authorization](../administration/authentication_and_authorization.html).
```
beckettsean: request*s*

beckettsean:
minor notes on Glossary, maybe a few things we should huddle on.

Really excited about this!

rkuchan:
Thank you for all the notes. I've made the requested changes (and a few more) to `key_concepts.md` and `glossary.md`.

beckettsean:
quick glance and all looks good. Let's :ship: it!

Issue #357, opened 2015-10-02
Batch Write Confusion


nathanielc:
When writing data using the HTTP API there is a timeout https://github.com/influxdb/influxdb/blob/master/etc/config.sample.toml#L86

This timeout applies to the entire write request as such it is necessary to batch writes to that they can complete within the time out. 

In the Writing data form a file section the here https://github.com/influxdb/influxdb.com/blob/master/content/docs/v0.9/guides/writing_data.md#writing-points-from-a-file

It explains how you can write data using curl and a file. I assumed that I could upload just any file, so I took my file with 5million lines and uploaded. Since that operation could not complete in the default timeout of `5s` I had to do some digging till I figured out that I should split up my file and upload it in batches. I recommend we add a note here explaining the timeout and how to batch your data so that it can be inserted within the timeout.

There is a command called `split` that I used. Might be worth mentioning.

 I ran the command `split -l5000 -d -a 5 mylargefile` and I ended up with ~1000 files with 5000 lines each. Then I used a simple for loop to upload them all. These details might not be helpful in the docs but thought I would add them here so you could see how I solved my issue.

PR #423, opened 2015-10-21
Add note to writing points from a file

rkuchan:
Fixes https://github.com/influxdb/influxdb.com/issues/357

### DIFF
--- a/content/docs/v0.9/guides/writing_data.md
+++ b/content/docs/v0.9/guides/writing_data.md
@@ -54,6 +54,8 @@ Write the data in `cpu_data.txt` to the `mydb` database with:
 <br>
 `curl -i -XPOST 'http://localhost:8086/write?db=mydb' --data-binary @cpu_data.txt`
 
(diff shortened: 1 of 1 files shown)


content/docs/v0.9/guides/writing_data.md
```diff
@@ -54,6 +54,8 @@ Write the data in `cpu_data.txt` to the `mydb` database with:
 <br>
 `curl -i -XPOST 'http://localhost:8086/write?db=mydb' --data-binary @cpu_data.txt`
 
+> **Note:** If your data file is big enough, it may be necessary to split that file into several files in order to write your data in batches to InfluxDB. [By default](https://github.com/influxdb/influxdb/blob/master/etc/config.sample.toml#L97), InfluxDB does not complete write operations that take longer than five seconds.
```
gunnaraasen: Lets say, "has more than 5000 points" instead of "big enough".

gunnaraasen: I may be wrong, but I think InfluxDB will still attempt to write the points even if the HTTP request times out at 5 seconds. There's just no confirmation they were written. Also, the timeout is configurable.

gunnaraasen:
Left a couple comments. Otherwise it looks good!

jackzampolin:
:+1:

PR #453, opened 2015-11-05
Added MD5 checksums to download page

rossmcdonald:
I'm assuming the content here is okay. My main concern is on the formatting of the page, since the checksums appear a little dis-jointed from the downloads themselves. 

Let me know if I should modify anything.

### DIFF
--- a/content/site/download.md
+++ b/content/site/download.md
@@ -21,12 +21,16 @@ layout = "sidebar"
 		wget https://s3.amazonaws.com/influxdb/influxdb_0.9.4.2_amd64.deb
 		sudo dpkg -i influxdb_0.9.4.2_amd64.deb
 
+MD5 checksum: `2f5c6a63d250568ef0d89bd8359dab79`
+
 #### RedHat & CentOS
 
 - 64-bit system install instructions
 
 		wget https://s3.amazonaws.com/influxdb/influxdb-0.9.4.2-1.x86_64.rpm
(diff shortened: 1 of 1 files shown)


otoolep:
+1 on the importance of this.

But tow is this going to work? Do our packaging scripts now spit our checksums? If not, I fear this will son become a pain to maintain.

rossmcdonald:
@beckettsean Checksum PR from our discussion yesterday. Does this look okay as far as the formatting goes?

rossmcdonald:
@otoolep Ideally this will be automated at some point, but right now we'll just have to manually update this on every new release. Releases don't seem to be super frequent, so I'm hoping it's not too much of a pain.

Also nightly builds will not be provided checksums until its automated.

otoolep:
How did you generate these checksums? Let's enhance `package.sh` to generate the checksums when the packages are generated, and dump the sums to stdout.

Sound good?

otoolep:
At least that way, the sums are there, ready to go.

rossmcdonald:
I just generated them all manually with `openssl dgst`, but I'll update the build script to spit them out on a packaging run.

otoolep:
Great, thanks. That way they are there, and always generated the same way.

gunnaraasen:
:+1:

beckettsean:
:+1:

Issue #276, opened 2015-09-02
Add non_negative_derivative to functions


pauldix:

beckettsean:
https://github.com/influxdb/influxdb/issues/1477

PR #455, opened 2015-11-06
edit and add to functions.md

rkuchan:
Format changes:

* Standardized the explanations of each function.
* Used the same sample data as the other InfluxQL docs for the examples.

Updates to content:

* `MIN()`, `MAX()`, `FIRST()`, `LAST()` now output the relevant timestamp instead of epoch 0 - added a note explaining InfluxDB's behavior in versions prior to 0.9.5.
* Made substantial edits to the explanation of `DERIVATIVE()`.
* Added discussions about `TOP()`, `BOTTOM()`, and `NON_NEGATIVE_DERIVATIVE()`.
* Added section about `AS` - we removed this from `data_exploration.md` and it should live somewhere.

`DERIVATIVE()`, `TOP()`, and `BOTTOM()` are new to me - any feedback on those sections would be particularly appreciated. When in doubt as to whether to include something about those functions I included it - the explanations may need some trimming.

Fixes https://github.com/influxdb/influxdb.com/issues/276

### DIFF
--- a/content/docs/v0.9/query_language/functions.md
+++ b/content/docs/v0.9/query_language/functions.md
@@ -523,7 +523,7 @@ SELECT MIN(<field_key>) FROM <measurement_name> [WHERE <stuff>] [GROUP BY <stuff
 
 Examples:
 
-* Select the mimimum `water_level` in the measurement `h2o_feet`:
+* Select the minimum `water_level` in the measurement `h2o_feet`:
 ```sql
 > SELECT MIN(water_level) FROM h2o_feet
 ```
@@ -730,7 +730,7 @@ Note that in the raw data, `water_level` equals `4.055` at `2015-08-18T04:06:00Z
 ## DERIVATIVE()
 Returns the rate of change for the values in a single [field](../concepts/glossary.html#field) in a [series](../concepts/glossary.html#series). InfluxDB calculates the difference between chronological non-`NAN` field values and converts those results into the rate of change per `unit`. The `unit` argument is optional and, if not specified, defaults to one second (`1s`).
 
-The basic `DERVATIVE()` query:
+The basic `DERIVATIVE()` query:
 ```sql
 SELECT DERIVATIVE(<field_key>, [<unit>]) FROM <measurement_name> [WHERE <stuff>]
 ```
@@ -745,7 +745,7 @@ Valid time specifications for `unit` are:
 
 `DERIVATIVE()` also works with a nested function coupled with a `GROUP BY time()` clause. For queries that include those options, InfluxDB first performs the aggregation, selection, or transformation across the time interval specified in the `GROUP BY time()` clause and then carries out the same procedure outlined above.
 
-The `DERVATIVE()` query with an aggregation function and `GROUP BY time()` clause:
+The `DERIVATIVE()` query with an aggregation function and `GROUP BY time()` clause:
 ```sql
 SELECT DERIVATIVE(AGGREGATION_FUNCTION(<field_key>),[<unit>]) FROM <measurement_name> WHERE <stuff> GROUP BY time(<aggregation_interval>)
 ```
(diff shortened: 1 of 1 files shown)


beckettsean:
I'm a little offput by the all caps "NULL". I'm used to seeing that as "null" in most or all contexts. Is there a stylistic reason to make it all caps?

content/docs/v0.9/query_language/functions.md
```diff
@@ -4,206 +4,1038 @@ aliases:
...
+2015-09-17T00:00:00Z	 3627.7619999999997
 ```
 
 # Selectors
 
-Selector functions return a single point from the range of points selected. Some selectors return all data for the point including timestamp. Some simply return the field value(s).
+## BOTTOM()
+Returns the smallest `N` values in a single [field](../concepts/glossary.html#field). The field type must be int64 or float64.
+```sql
+SELECT BOTTOM(<field_key>[,<tag_keys>],<N>)[,<tag_keys>] FROM <measurement_name> [WHERE <stuff>] [GROUP BY <stuff>]
```
beckettsean: Can a user also select another field_key in the SELECT statement, or does that throw an error?

content/docs/v0.9/query_language/functions.md
```diff
@@ -4,206 +4,1038 @@ aliases:
...
+```sh
+name: h2o_feet
+--------------
+time			               bottom	 location
+2015-08-29T14:30:00Z	 -0.61	  coyote_creek
+2015-08-29T14:36:00Z	 -0.591	 coyote_creek
+2015-08-30T15:18:00Z	 -0.594	 coyote_creek
+```
+
+* Select the smallest value of `water_level` within each tag value of `location`:
```
beckettsean: might need to contrast this with `SELECT BOTTOM(water_level,2) FROM h2o_feet GROUP BY location`.

how does having `location` in the `BOTTOM` function differ from just having it outside the function?

content/docs/v0.9/query_language/functions.md
```diff
@@ -4,206 +4,1038 @@ aliases:
...
+2015-09-02T00:00:00Z	 10451.321000000013
+2015-09-07T00:00:00Z	 10871.817999999988
+2015-09-12T00:00:00Z	 11459.001000000007
+2015-09-17T00:00:00Z	 3627.7619999999997
 ```
 
 # Selectors
 
-Selector functions return a single point from the range of points selected. Some selectors return all data for the point including timestamp. Some simply return the field value(s).
+## BOTTOM()
```
beckettsean: Only wired up in 0.9.5+, we should note that.

content/docs/v0.9/query_language/functions.md
```diff
@@ -4,206 +4,1038 @@ aliases:
...
+
+name: h2o_feet
+tags: location = santa_monica
+time			               first
+----			               -----
+1970-01-01T00:00:00Z	 2.064
+```
+
+## LAST()
+Returns the newest value (determined by the timestamp) of a single [field](../concepts/glossary.html#field).
```
beckettsean: Might be worth a note that LAST won't return points in the future unless the WHERE clause explicitly includes that time range.

```
> select * from ltest where time < now() + 100w
name: ltest
-----------
time			value
1447112210023536344	12
1457112184000000000	14

# second point is in the future

> select last(value) from ltest
name: ltest
-----------
time			last
1447112210023536344	12

> select last(value) from ltest where time < now() + 100w
name: ltest
-----------
time			last
1457112184000000000	14
```

content/docs/v0.9/query_language/functions.md
```diff
@@ -4,206 +4,1038 @@ aliases:
...
+
+<dt> See GitHub Issue [#3691](https://github.com/influxdb/influxdb/issues/3691) for more information. </dt>
+
+## HISTOGRAM()
+`HISTOGRAM()` is not yet functional. 
+
+<dt> See GitHub Issue [#3674](https://github.com/influxdb/influxdb/issues/3674) for more information. </dt>
 
+## NON_NEGATIVE_DERIVATIVE()
+Returns the non-negative rate of change for the values in a single [field](../concepts/glossary.html#field) in a [series](../concepts/glossary.html#series). InfluxDB calculates the difference between chronological non-`NAN` field values and converts those results into the rate of change per `unit`. The `unit` argument is optional and, if not specified, defaults to one second (`1s`).
```
beckettsean: I prefer `NaN` as the spelling, rather than `NAN`

rkuchan:
Made the requested changes!

beckettsean:
:+1: from me!

PR #470, opened 2015-11-12
Correct escaping rules documentation

### LINKED ISSUE #469: "Escaping Characters" incorrectly describes behaviour for measurements
Hi,

From the [Write Syntax](https://influxdb.com/docs/v0.9/write_protocols/write_syntax.html) page:

> **Escaping Characters**
> If a measurement, tag key, tag value, or field key contains a space , comma `,`, or an equals sign `=` it must be escaped using the backslash character `\`. Backslash characters do not need to be escaped.

Yet the current behaviour is that measurements do not need to have `=` characters escaped. For example:

```
> insert e=mc2,tag=val value=1
> select * from "e=mc2"
name: e=mc2
-----------
time			tag	value
1447345080888092606	val	1
```

e-dard:
Correct escaping rules for measurements.

### DIFF
--- a/content/docs/v0.9/write_protocols/write_syntax.md
+++ b/content/docs/v0.9/write_protocols/write_syntax.md
@@ -54,7 +54,7 @@ Tag keys and values, and field keys and values must be separated by the equals s
 
 ### Escaping Characters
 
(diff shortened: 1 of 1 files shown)


e-dard:
Fixes #469

gunnaraasen:
Thanks @e-dard. Can you please sign the [CLA](https://influxdb.com/community/cla.html)?

e-dard:
@gunnaraasen done!

gunnaraasen:
Thanks again!

PR #471, opened 2015-11-12
Kapacitor updates

nathanielc:
Adds a How to Contribute to Kapacitor page.
New Leaderboard use case.
Lots of update to the TICKscript node docs.


nathanielc:
@gunnaraasen Could you take a look at this PR? Its just a bunmch of updates to the Kapacitor docs since things have changed.

content/docs/kapacitor/v0.1/tick/eval_node.md
```diff
@@ -26,22 +30,34 @@ Property methods modify state on the calling node. They do not add another node
...
 If called the existing fields will be preserved in addition 
-to the new field being set. 
+to the new fields being set. 
+
+Optionally a list of field names can be given. 
+Only fields in the list will be kept. 
+If no list is given then all fields are kept. 
+The list of fields to keep is evaluated after all 
+expressions have been evaluated. 
+This way intermediate values can be discarded.
```
gunnaraasen: Should want to lead with this point.

content/docs/kapacitor/v0.1/tick/join_node.md
```diff
@@ -3,21 +3,33 @@ title: JoinNode
...
 ---
 
-Joins the data from two nodes in the order that data arrives. 
+Joins the data from any number of nodes in the order that data arrives. 
+The join can be an inner or outer join, see the [JoinNode.Fill](/docs/kapacitor/v0.1/tick/join_node.html#fill) property. 
 As each data point is received from a parent node it is paired 
-with the next data point from the other parent node. 
+with the next data point from the other parent node with a 
+matching timestamp. 
+A tolerance can be supplied in order to consider points within the give
```
gunnaraasen: give -> given

content/docs/kapacitor/v0.1/tick/join_node.md
```diff
@@ -3,21 +3,33 @@ title: JoinNode
...
 
-Joins the data from two nodes in the order that data arrives. 
+Joins the data from any number of nodes in the order that data arrives. 
+The join can be an inner or outer join, see the [JoinNode.Fill](/docs/kapacitor/v0.1/tick/join_node.html#fill) property. 
 As each data point is received from a parent node it is paired 
-with the next data point from the other parent node. 
+with the next data point from the other parent node with a 
+matching timestamp. 
+A tolerance can be supplied in order to consider points within the give 
+tolerance duration will be considered to be the same time.
```
gunnaraasen: > A tolerance can be supplied in order to consider points within the given tolerance duration will be considered to be the same time.

should be

> A tolerance can be supplied to join points that do not have perfectly aligned timestamps. Any points that fall within the tolerance are joined on the timestamp.

Also the behavior seems a bit ambiguous. What if there are multiple points within the tolerance duration, does it only take the first point or all points? Is the tolerance duration +/- or only + from the point's timestamp?

nathanielc: Does this make sense?
> If multiple points fall within the tolerance window than they are joined in the order
they arrive one by one.

content/docs/kapacitor/v0.1/tick/stream_node.md
```diff
@@ -31,6 +31,27 @@ Properties
 
 Property methods modify state on the calling node. They do not add another node to the pipeline and always return a reference to the calling node.
 
+### Truncate
```
nathanielc: Its the name Go used so I borrowed it.  Hmm, mathematically the value is literally being truncated where all digits below the truncate value are discarded. Can't think of a better one either.

content/docs/kapacitor/v0.1/use_cases/index.md
```diff
@@ -0,0 +1,19 @@
+---
+title: Use Cases
```
gunnaraasen: I think this section would be better named as "Guides", "Examples", or "Tutorials".

nathanielc:
@gunnaraasen Thanks for the feedback!

How do the docs for the JoinNode look now?

content/docs/kapacitor/v0.1/tick/stream_node.md
gunnaraasen: Ok, I can't think of anything better either.

gunnaraasen:
:+1: The `JoinNode` description is much better. Thanks! This looks good to merge.

content/docs/kapacitor/v0.1/tick/eval_node.md
```diff
@@ -26,22 +30,34 @@ Property methods modify state on the calling node. They do not add another node
 
 ### As
 
-The name of the field that results from applying the expression. 
+List of names for each expression.
```
nathanielc: Added examples with more than one expression for both the As and Keep properties.

gunnaraasen: Thanks! Makes sense now. For some reason I was thinking multiple expressions == multiple `eval`s.

PR #478, opened 2015-11-13
typo fix when subtracting time & comment removal

kostasb:
Changed line 125 from "SELECT * FROM foodships WHERE time > now() -1h" to "SELECT * FROM foodships WHERE time > now() - 1h" as it doesn't work without a whitespace after the '-' character and might cause confusion.

Also removed "a function in progress" comment for DERIVATIVE as the function has been implemented and documented.


rossmcdonald:
+1, good catch

beckettsean:
@kostasb Looks good! Merge the PR whenever you like.

PR #489, opened 2015-11-16
removed references to POST with --data-urlencode

kostasb:
Removed misleading references of curl commands using POST method while passing attributes with data-urlencode. This does not apply, as the string defined in data-urlencode when using a POST method is passed to the content instead of getting appended to the URL.

Troubleshooting and justification:
---
POST method with attributes defined in data-urlencode section. The string is not appended to the POST URL, rather it is added in the content along with the measurement data.

curl -vvvX POST 'http://localhost:8089/write' --data-urlencode 'db=mydb' --data-binary 'disk_free,hostname=server01 value=442221834240i'

[kostas@localhost ~]$ sudo nc -l -p 8089
POST /write HTTP/1.1
User-Agent: curl/7.29.0
Host: localhost:8089
Accept: */*
Content-Length: 55
Content-Type: application/x-www-form-urlencoded

db=mydb&disk_free,hostname=server01 value=442221834240i

---
POST method using attributes originally appended to the URL which does work:

curl -vvvX POST 'http://localhost:8089/write?db=mydb' --data-binary 'disk_free,hostname=server01 value=442221834240i'

[kostas@localhost ~]$ sudo nc -l -p 8089
POST /write?db=mydb HTTP/1.1
User-Agent: curl/7.29.0
Host: localhost:8089
Accept: */*
Content-Length: 47
Content-Type: application/x-www-form-urlencoded

disk_free,hostname=server01 value=442221834240i

---
GET method appends data-urlencode content to the URL but can't be used to write data with the HTTP API.

curl -G 'http://localhost:8089/write' --data-urlencode 'db=mydb' --data-binary 'disk_free,hostname=server01 value=442221834240i'

[kostas@localhost ~]$ sudo nc -l -p 8089
GET /write?db=mydb&disk_free,hostname=server01 value=442221834240i HTTP/1.1
User-Agent: curl/7.29.0
Host: localhost:8089
Accept: */*

### DIFF
--- a/content/docs/v0.9/write_protocols/write_syntax.md
+++ b/content/docs/v0.9/write_protocols/write_syntax.md
@@ -230,7 +230,7 @@ curl -X POST 'http://localhost:8086/write?db=mydb&precision=ms' --data-binary 'd
 
 You can also pass a file using the `@` flag. The file can contain a batch of points, one per line. Points must be separated by newline characters `\n`. Batches should be 5000 points or fewer for best performance.
 
-`curl -X POST 'http://<hostname>:<port>/write?db=<database>' --data-binary @<filename>`
+`curl -X POST 'http://localhost:8086/write?db=<database>' --data-binary @<filename>`
 
 ### Caveats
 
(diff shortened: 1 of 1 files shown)


rossmcdonald:
To add to your message above, here is a relevant snippet from the `curl` man page:

> When used, this option will make all data specified with -d, --data, --data-binary or --data-urlencode to be used in an HTTP GET request instead of the POST request that otherwise would be used. The data will be appended to the URL with a '?' separator.

http://curl.haxx.se/docs/manpage.html#-G

Great catch, though. +1

beckettsean:
:+1:  merge it

PR #492, opened 2015-11-17
Bump telegraf version to 0.2.1

sparrc:
fyi: homebrew PR is pending


sparrc:
going straight to 0.2.2 instead #495

PR #496, opened 2015-11-18
Add anchors for each product on downloads page

wfro:
As I was writing a release announcement for Chronograf I found myself wanting to link directly to a specific section on the downloads page.

This PR adds anchors which can be linked to by adding `#influxdb`, `#telegraf`, or `#chronograf` when linking to the downloads page:

gif madness

Top level navigation linking to each anchor might be nice as well in the future.


gunnaraasen:
The current download page already has anchors that work like this, e.g. https://influxdb.com/download/#telegraf. Making the headers themselves linkable would be nice though.

wfro:
Ah ok, thanks!  Closing.

PR #503, opened 2015-11-19
update docs for new from format

nathanielc:
Changed the format of the from statement to be explicit so that you don't have to worry about weird double parsing errors. See https://github.com/influxdb/kapacitor/pull/59


rkuchan:
@nathanielc, this looks good to me!

nathanielc:
thanks @rkuchan

PR #504, opened 2015-11-19
Add an extended discussion about the CQ configuration options

rkuchan:
`extended_cqconfig_options.md` offers an in-depth discussion about the continuous query (CQ) configuration options and how they affect different CQs.

All feedback is welcome - the wording in this document has been a particular challenge.

### DIFF
--- a/content/docs/v0.9/query_language/continuous_queries_config.md
+++ b/content/docs/v0.9/query_language/continuous_queries_config.md
@@ -1,0 +1,158 @@
+---
+title: Configuring continuous queries
+---
+
+A continuous query (CQ) is an automated query that executes over recent time intervals. When InfluxDB executes a CQ it computes the query for the current `GROUP BY time()` interval (determined by `now()`) and it computes the query for previous `GROUP BY time()` intervals.
+
+The rate at which InfluxDB executes a CQ and the number of previous time intervals that InfluxDB queries depend on the CQ's `GROUP BY time()` interval and on the following configuration options (shown with their default settings):
+
+* `compute-runs-per-interval = 10`
+* `compute-no-more-than = "2m0s"`
+* `recompute-previous-n =  2`
+* `recompute-no-older-than = "10m0s"`
+
+InfluxDB compares `compute-runs-per-interval` and `compute-no-more-than` to determine the rate at which it executes a CQ. Under the first setting, InfluxDB executes the CQ `10` times per the `GROUP BY time()` interval. Under the second setting, InfluxDB executes the CQ every `2` minutes; this is the minimum rate at which InfluxDB executes a CQ.
+
+InfluxDB compares `recompute-previous-n` and `recompute-no-older-than` to determine the number of previous intervals it queries when it executes a CQ. Under the first setting, InfluxDB queries `2` previous intervals. Under the second setting, InfluxDB queries all previous intervals within `now()` - `10` minutes. InfluxDB queries the smaller of the two options when it executes a CQ.
+
+Depending on your CQ's `GROUP BY time()` interval you may want to make changes to the default configuration settings. The next sections present example CQs and discuss how they work with the configuration options:
+
+* [CQ 1](../query_language/continuous_queries_config.html#cq-1-4-minute-group-by-time-interval) is an introductory example CQ that shows how the default configuration settings work with a CQ's `GROUP BY time()` interval. 
+* [CQ 2](../query_language/continuous_queries_config.html#cq-2-30-second-group-by-time-interval) is an example of a CQ with a shorter term `GROUP BY time()` interval that doesn't work well with the default configuration settings. This section also covers how to alter the configuration settings to fix the issue.
+* [CQ 3](../query_language/continuous_queries_config.html#cq-3-1-hour-group-by-time-interval) is an example of a CQ with a longer term `GROUP BY time()` interval. This section shows how to alter the configuration settings to take into account lagged data.
+
+## CQ 1: 4 minute `GROUP BY time()` interval
+### The default CQ schedule
+Default configuration settings:  
+`compute-runs-per-interval = 10`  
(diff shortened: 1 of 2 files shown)


content/docs/v0.9/query_language/continuous_queries_config.md
```diff
@@ -0,0 +1,158 @@
+---
+title: Configuring continuous queries
+---
+
+A continuous query (CQ) is an automated query that executes over recent time intervals. When InfluxDB executes a CQ it computes the query for the current `GROUP BY time()` interval (determined by `now()`) and it computes the query for previous `GROUP BY time()` intervals.
```
beckettsean: Trying to clarify, use what helps:

A continuous query (CQ) is a normal query that the system periodically executes against recent time intervals. When InfluxDB creates a batch of CQs it generates a query where `now()` is between the beginning and end of the ``GROUP BY time()` interval as well as generating the query for some number of recent `GROUP BY time()` intervals.

content/docs/v0.9/query_language/continuous_queries_config.md
```diff
@@ -0,0 +1,158 @@
...
+
+The rate at which InfluxDB executes a CQ and the number of previous time intervals that InfluxDB queries depend on the CQ's `GROUP BY time()` interval and on the following configuration options (shown with their default settings):
```
beckettsean: Same note:

The rate at which InfluxDB generates batches of CQs and the number of queries generated per batch depends on the CQ's `GROUP BY time()` interval and on the following configuration options (shown with their default settings):

content/docs/v0.9/query_language/continuous_queries_config.md
```diff
@@ -0,0 +1,158 @@
...
+
+* `compute-runs-per-interval = 10`
+* `compute-no-more-than = "2m0s"`
+* `recompute-previous-n =  2`
+* `recompute-no-older-than = "10m0s"`
+
+InfluxDB compares `compute-runs-per-interval` and `compute-no-more-than` to determine the rate at which it executes a CQ. Under the first setting, InfluxDB executes the CQ `10` times per the `GROUP BY time()` interval. Under the second setting, InfluxDB executes the CQ every `2` minutes; this is the minimum rate at which InfluxDB executes a CQ.
```
beckettsean: InfluxDB compares `compute-runs-per-interval` and `compute-no-more-than` to determine the rate at which it generates batches for a given CQ. The `compute-runs-per-interval` setting is an upper bound on the number of incremental queries generated per `GROUP BY time()` interval. The `compute-no-more-than` setting is the maximum frequency with which batches will be generated. `compute-no-more-than` takes precedence over `compute-runs-per-interval`.

As set above, InfluxDB will generate batches for each CQ as often as `10` times per the `GROUP BY time()` interval. However, batches will not be generated any more often than every two minutes, regardless of the value if  `GROUP BY time()` divided by 10. 

For example, if the CQ has a `GROUP BY time()` interval of `30m`, then the `compute-runs-per-interval` setting sets a maximum batch frequency of `3m`, since `30m` / 10 = `3m`. This is less frequent than the maximum frequency of `2m`, as set by `compute-no-more-than`, so the system will generate a batch of that CQ every `3m`. 

If the CQ has a `GROUP BY time()` interval of `30s` then the `compute-runs-per-interval` setting implies that batches will be generated every `3s`, since `30s` / 10 = `3s`. However, the `compute-no-more-than` setting takes precedence and thus batches will only be generated every `2m`.

content/docs/v0.9/query_language/continuous_queries_config.md
beckettsean: The generated queries have time ranges defined on round-number boundaries, i.e. given `WHERE time >= timestamp1 AND time =< timestamp2`, `timestamp` % 60 and `timestamp2` % 60 will both equal zero for any query spanning at least one minute.

content/docs/v0.9/query_language/continuous_queries_config.md
```diff
@@ -0,0 +1,158 @@
...
+
+InfluxDB compares `recompute-previous-n` and `recompute-no-older-than` to determine the number of previous intervals it queries when it executes a CQ. Under the first setting, InfluxDB queries `2` previous intervals. Under the second setting, InfluxDB queries all previous intervals within `now()` - `10` minutes. InfluxDB queries the smaller of the two options when it executes a CQ.
```
beckettsean: argh! lost my prior edits :(

### queries per batch

InfluxDB uses `recompute-previous-n` and `recompute-no-older-than` to determine the number of queries generated per batch for a given CQ. All batches will include at least one generated query with the time range spanning `now()`. The maximum number of queries spanning intervals prior to `now()` is determined by `recompute-previous-n`. Therefore (`recompute-previous-n` + 1) is the maximum number of queries per batch. `recompute-no-older-than` puts a limit on the age of the previous query interval, such that the upper time boundary of each generated query must be greater than `now()` - `recompute-no-older-than`. `recompute-no-older-than` takes precedence over `recompute-previous-n`.

beckettsean: As set above, InfluxDB will generate at most three queries per batch: one spanning `now()` and up to two queries spanning older time intervals, as set by `recompute-previous-n`. In addition, the upper boundary of the query time range for each generated query must be less than `10m` in the past.

For example, if the CQ has a `GROUP BY time()` of `30s`, then three queries will be generated for each batch. However, if the CQ has a `GROUP BY time()` of `30m`, then only one or two queries will be generated for a given batch, since the third query would have an upper time boundary at least `30m` in the past, which exceeds the `recompute-no-older-than` setting.

beckettsean:
@rkuchan is this still active or is there a replacement on docs.influxdata.com?

rkuchan:
Closed in favor of: https://github.com/influxdb/docs.influxdata.com/pull/93

PR #524, opened 2015-11-24
update downloads for Kapacitor

nathanielc:
NOTE: These download links and MD5Sums are not correct yet. Still need to build and publish final binary for release. I'll update this PR once its ready.


gunnaraasen:
Structure looks correct. Might want to remove the extra zero (`0.1.1.0`) in the download links. Not sure if that's intentional.

PR #534, opened 2015-12-01
Telegraf 0.2.3 version bump

sparrc:


rossmcdonald:
Updated the hashes and fixed a typo in the tar.gz URL.

PR #538, opened 2015-12-04
update current version to 0.9.5.1

ORANGE-XFM:


beckettsean:
Good catch, @ORANGE-XFM, thanks!

Before we can merge, can you sign the [CLA](https://influxdb.com/community/cla.html)?

ORANGE-XFM:
I can't currently (google docs is blocked here), so please ignore this CR and change it directly

